PR #1, opened 2017-09-29
Changing normalisation coefficient in del-tau for BOG and geen

sit23:
Discussion that using dtau/dsigma was a bad way to go, rather than using dtau/dp. This is because dtau/dsigma would imply that optical depths at the surface over a mountain (say at 300hPa) would be the same as optical depths at the surface over flat land (say at 1000hPa) as both would correspond to sigma = 1. 

What we really want is the optical depth to depend on the amount of gas along the line of sight, so it should be related to pressure, rather than sigma. Suggestion was to follow Frierson lead and to use dtau/dp, but to normalise pressure by a constant, which gives the optical depths in simple units, but still retains the dtau/dp structure.


jamesp:
Sounds good stephen.  On a slightly related note, do we use the `pstd` and `pstd_mks` (as far as I can tell the difference is in the units) in different places in the code? Potentially seems like a place to inadvertently introduce a bug when changing the value of one but not the other.

sit23:
Thanks @jamesp. The `pstd` constant is not actually used anywhere in Isca, with `pstd_mks` being used within the two-stream code and in the initialisation of the pressure variables. `Pstd` is left over from this variable being used in aerosol and cloud schemes in more complicated GFDL configurations (see e.g. https://github.com/mjucker/MiMA/blob/master/src/atmos_param/sea_esf_rad/gas_tf.f90). Isca compiles and runs normally if `pstd` is removed from `constants.f90`, so we could just remove it to avoid confusion?

sit23:
I have now updated this P/R so that the normalisation of lw coefficients is done by `pstd_mks_earth`, which is a new parameter in the `constants.f90` file that should not be changed. The surface pressure, set by `pstd_mks` is changeable via a nml entry, and can be different to `pstd_mks_earth`, but Byrne and Geen lw schemes will still keep the same lw absorption coefficients. This is because they are defined relative to actual absorption data, which will not change with surface pressure.

sit23:
Having discussed with @jamesp, have removed `reference_sea_level_press` as a namelist option from spectral_dynamics, and have made it equal to pstd_mks. That way, there is only one namelist variable required to change surface pressure, and it's pstd-mks.

RuthG:
This all sounds sensible to me, though it does mean future runs may have slight differences anywhere pstd-mks is carried through I guess. I don't expect these will be significant though, so am happy for it to be merged when you're ready.

PR #3, opened 2017-10-17
New python framework

jamesp:
This is the second iteration on the Python framework.  It's only been tested by me so far, so probably not suitable to go in the Isca1.0 release.

This reworks the system, functionally separating GFDL_BASE from the experiment so that python experiment files should now be truly standalone and individually reproducible.  

There is a README in `exp/test_cases_py` and a couple of example python scripts `held_suarez.py` and `example_basic.py` that show some of the changes.

A lot of optional features that had been put into the Experiment class are now in a `isca.util` module.

Now there are two, pared down classes:

CodeBase: Handles which version of the code you are using, and compilation.
Experiment: Handles configuration and running the model.

See `exp/test_cases_py` for examples.


jamesp:
I think we should try and pull this into master this week if possible. It is very likely there are going to be lots of bugs - it would be good to get everyone who uses the Python scripts to give it a go and report back to me with what doesn't work

sit23:
Sounds good to me. I guess a good starting point once it's in would be to try and get all the test cases I made running under the new framework, including making a giant-planet codebase object.

jamesp:
I'll work on the test cases today/tomorrow.  I think I might need to talk with @gregcolyer and @Alexander-P to make sure their recent PRs don't conflict directly with the changes

sit23:
Happy to help with test cases if you like, particularly as there are quite a few!

jamesp:
I've updated two of the test cases to work with the new framework: Held Suarez and MiMA.  Both now run, but I haven't validated output.  @sit23 you may want to add more namelist values to the MiMA test case now that it the python scripts are not populating the namelist from `core.nml` and `phys.nml`.

sit23:
Will do - thanks @jamesp.

PR #4, opened 2017-10-19
Adding Isca test cases and q-flux calculation scripts

sit23:
I have added several test cases for Isca, and have removed many of the old example experiments. The README in `Isca/exp/test_cases/README.md` provides information on each of the test cases, and on the location and use of many of the python scripts in `src/extra/python/scripts/`. Each of the test cases has been run and they all appear to be stable.

Other big change is I have added the `calculate_qflux.py` script and associated routines. This was previously part of my branch on ExeClim/ShareCode. All scripts have been tested and are functional.


jamesp:
This is brilliant @sit23.  It should be very straightforward to convert these to run on the new python framework in the 2.0 too.

sit23:
Following @gregcolyer's suggestion, I have removed repeated copies of the `ozone_1990.nc` file, and put it in the now streamlined `Isca/input` folder, from which I have also removed lots of old files. The test cases involving RRTM now all point directly to the single `ozone_1990.nc` file, without using any symlinks. I have done the same with the `era_land_t42.nc` file.

PR #7, opened 2017-10-24
Edits to allow zonally symmetric running 

RuthG:
Edits to 4 files to pass namelist variable 'make_symmetric' through to spherical.F90
I have used local variables echoing how the code currently handles these namelist variables that are passed through several files (e.g. south_to_north_local) but am happy to take input/advice on this.


sit23:
Looks good to me!

RuthG:
I've added a test case using the set up I've been using for my axisymmetric runs, with a bit of an explanation of the options used. Any other comments welcome, and if anyone wants to test it all out that'd be great.

PR #8, opened 2017-10-24
Top-down forcing changes

Alexander-P:
A new equilibrium temperature scheme added to hs_forcing with necessary hooks.
Extra function in hs_forcing, everything should be still switched with the equilibrium_t_option namelist variable. hs_forcing_init will do the required spin up if this toggle is set.
Changes to atmosphere to feed through extra variables for top-down to work.
A couple of variables added to the astro and constants namelist to allow their use in hs_forcing.


sit23:
This looks good to me @Alexander-P - nice work! If you're happy it works, then I'm happy too. Perhaps the only suggestion I would make would be for your test experiment to go in a folder within the `test_cases` bit. E.g. `Isca/exp/test_cases/top_down_relaxation`?

RuthG:
Agree, looks good, excited to use it!

PR #9, opened 2017-10-24
Implementation of Relaxed Arakawa-Schubert Scheme

sit23:
@penmaher added the Relaxed Arakawa-Schubert scheme to the `ExeClim/GFDLmoistModel` repo, but the branch it was on (`RAS_conv`) wasn't transferred to Isca. I copied those changes, and included a few tweaks, and it's now running within Isca. Running a test case at the moment, but I think it should go in v1.0, given that the paper says `A simplified mass flux scheme (relaxed Arakawa– Schubert) can be incorporated but has not been tested.`

### DIFF
--- a/src/atmos_param/ras/moist_processes.f90
+++ b/src/atmos_param/ras/moist_processes.f90
@@ -1,0 +1,2851 @@
+
+                    module moist_processes_mod
+
+!-----------------------------------------------------------------------
+!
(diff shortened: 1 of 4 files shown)


RuthG:
Edits to idealized_moist look ok to me, it'll be interesting to see the tests. Happy for it to be merged if the tests look alright.

PR #11, opened 2017-10-26
Bring two_stream code in line with Isca paper

RuthG:
Updates to LW non-window scheme in the two stream code to the version described in the paper.


sit23:
Looks good to me @RuthG

PR #13, opened 2017-10-31
Edits to constants.F90 to add option that seconds_per_sol is a simple…

RuthG:
… scaling of 86400 with the rotation rate to be used


sit23:
Looks good to me.

PR #14, opened 2017-10-31
Bug fix in mixed_layer.F90 to add '.nc' to non time varying qflux inp…

RuthG:
…ut file

### DIFF
--- a/src/atmos_spectral/driver/solo/mixed_layer.F90
+++ b/src/atmos_spectral/driver/solo/mixed_layer.F90
@@ -373,11 +373,11 @@ if (load_qflux) then
 	   call interpolator_init( qflux_interp, trim(qflux_file_name)//'.nc', rad_lonb_2d, rad_latb_2d, data_out_of_bounds=(/CONSTANT/) )
 	else
 
-	   if(file_exist(trim(qflux_file_name))) then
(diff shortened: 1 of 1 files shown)


sit23:
Thanks for this @RuthG - looks good to me.

Just as note to myself, time-varying input-file names using the interpolator don't need the `INPUT/` at the front of their filename, as the interpolator adds this automatically. But when reading non-time-varying files the `INPUT/` bit is necessary as it's not assumed by the non-interpolator file-reading method. This applies for this non-time varying qflux and the land input file.

jamesp:
This INPUT/ no INPUT sounds like something we should pick one and standardise

sit23:
Indeed. There's a standardisation issue as far as the lines of code in the fortran are concerned, which is what I was referring to above, but as far as the user is concerned, @RuthG's change has standardised what's required for the q-flux file names. The namelist option you pass is now just `name_of_file` for both the time-constant and time-varying q-fluxes, which it wasn't before. Before @RuthG's  change, the time-constant input namelist option would have needed to be `INPUT/name_of_file.nc` where the time-varying one would just have been `name_of_file`. @RuthG has fixed that inconsistency here. The only place this need for `INPUT/` remains is the land-file input. I can P/R the appropriate change so that it's all consistent as far as the user is concerned.

PR #16, opened 2017-11-14
Updates to test cases in new python framework

sit23:
All test cases have been updated to the new python framework, and have been tested for bitwise reproducibility compared with same experiment under old python framework.


jamesp:
Looks great @sit23, thanks for doing all of this

PR #17, opened 2017-11-14
Updated test_cases/top_down_test

jamesp:
- top_down_test now uses the new isca python framework
- parameter_sweep example moved to held_suarez_test folder
- removed `test_cases_py` folder as has been superseded by more comprehensive `test_cases`


sit23:
This looks great @jamesp. Only thing is it would be great to still have a copy of `isca_job.sh` somewhere. Is your feeling that this kind of Exeter-specific file doesn't belong in the repo? If not, I'll download the file and am happy for it to come out of the repo.

jamesp:
Good point - I think it's probably worth salvaging.  It could maybe go in the `test_cases` folder as an example of how any of the codes could be run from a batch job script

sit23:
Yes, I think if it goes in there as an example, then that's definitely helpful for more than just Exeter users.

RuthG:
Looks good to me :)

sit23:
Thanks @jamesp - will merge this in now.

PR #18, opened 2017-11-15
Minor edit to derive method in Experiment

RuthG:
I’ve found a very little problem in the updated python scripts when using ‘derive’, it tries to apply .copy() to self.inputfiles, but this is a list and does not have a copy method and so it brings up an error message. Commenting out .copy() runs fine.


jamesp:
I don't use input files so don't know what the desired behaviour is here, it's probably best if you choose, but something to consider:

Without a copy, both original experiment and any derived experiments will point to the exact same list of input files.  So that if you e.g. 
```exp=baseexp.derive('new1')
exp.inputfiles.append('newfile.nc')```
 'newfile.nc' will also get appended to the `baseexp.inputfiles`.  So that means another call to `baseexp.derive('new2')`, this second derived experiment will also now have 'newfile.nc' in it's inputfiles.  I can see this having unintended consequences in a parameter sweep where you might want to append a different input file for each point in the sweep.

If you don't want that to be the case, then replacing `.copy()` with `[:]` should mean that each new experiment gets it's own list object for inputfiles.

I'm mostly agnostic, but would lean towards making a copy for each experiment object as it feels more obvious semantics.  Really it depends how the people who actually use the input files want it to work though!

RuthG:
That makes sense, thanks! It was really only a hack to get rid of the error message from having .copy() attached to the list object. If it'll make a new object, I'll add a [:] once I'm back on that branch.

sit23:
This looks good to me. Will give it a try on one of my experiments with an input file.

RuthG:
I am merging this as no-one seems to have any objections.

PR #19, opened 2017-11-15
Move some shared parameters from test cases to a separate file.

RuthG:
This is less intended as a pull request to close, more to help in discussion.
I have created a file in exp/test_cases with the defaults from Stephen's python snippet, which has the defaults from phys.nml and core.nml. I've added the Frierson levels on the bottom so these are easy to implement too. I've also edited the test cases* so they read this in, then modify any parameters that need to be changed for the experiment.

From the amount of pink in the file comparisons, this removes many lines of code. I'm happy to be overruled, but I think this makes it simpler to see which are relevant options for each test case. I'm assuming here that the purpose of the test cases are to get new users started, as well as to test our commits haven't broken anything. 

NB I'm just testing that these scripts run for the time being. I can do a proper check that the test case results are unchanged if people want to go down this route.

Of course, the best thing would be to eliminate the need for a file like this altogether. We could also discuss whether we'd be happy to put some of these into the source code.

--------
* Except for top_down, which still seems to use the old python code. I didn't want to cause complications in case there is a reason for this! :)


sit23:
Hey @RuthG - thanks very much for doing this. I very much appreciate the time you've put into sorting it out in all those test cases! :) 

Having had some time to think about it, I'm now in favour of doing this. I think the main problem with the template namelists, as you said, was that they were hidden / you wouldn't know they were there unless you dug through the code. The method proposed above gets rid of that problem, and comes with the advantages of:
* Making the exp files shorter and easier to read
* Will make the process of updating old experiments to the new python library easier
* Gives people the option of not using this template file, meaning they can have all their values set inside one python script if they so choose.

I know I originally said I wasn't in favour of this to you @jamesp, but having changed my mind, I'm wondering what do you think?

gregcolyer:
I notice there are some duplicate commits in the above list, e.g. 
d85790b looks the same as 37c8d08 (which latter is already merged into master). Does this matter? It may well not.

jamesp:
I understand the desire to do this as it solves the problem of recreating these "default" parameters for each experiment that you run.  My two concerns can be summarised as:

* What's default to you is not default to me.  I spent a long time (too long) at the start of my PhD trying to get the model to run with the default MiMa values but with exoplanet parameters mixed in.  It was very frustrating and I believe I would have been better served *not* starting from this baseline.
* If you want to guarantee bitwise reproduction of results of your experiments in the future, version control is more difficult when namelist and and experiment files are separate.  If you decide that your "default" namelist should be slightly different in 12 months time, you should consider that now rerunning any of your old experiments that import it will give different results.

That said, I think this is the way most people will setup their own workflow, and it's likely that most people will be running the model as an Earth configuration.  So an example namelist of this form would be well received.

My objections are only around the unintended/unexpected implications of this separation.  Given that I've now documented them here for future laughs, I'll happily go along with the consensus view.

gkvallis:
Maybe we need some face-to-face discussion over this? We were thinking of having it at the team meeting but we missed that this morning, and in any case it just needs a subgroup. Stephen, Ruth, Jamie and anyone else who is interested? Could be tomorrow (Thursday) morning at 9.15 over coffee, or your drug of choice, or over tea in the afternoon.

-geoff


On 22 Nov 2017, at 12:42 pm, James Penn <notifications@github.com<mailto:notifications@github.com>> wrote:


I understand the desire to do this as it solves the problem of recreating these "default" parameters for each experiment that you run. My two concerns can be summarised as:

  *   What's default to you is not default to me. I spent a long time (too long) at the start of my PhD trying to get the model to run with the default MiMa values but with exoplanet parameters mixed in. It was very frustrating and I believe I would have been better served not starting from this baseline.
  *   If you want to guarantee bitwise reproduction of results of your experiments in the future, version control is more difficult when namelist and and experiment files are separate. If you decide that your "default" namelist should be slightly different in 12 months time, you should consider that now rerunning any of your old experiments that import it will give different results.

That said, I think this is the way most people will setup their own workflow, and it's likely that most people will be running the model as an Earth configuration. So an example namelist of this form would be well received.

My objections are only around the unintended/unexpected implications of this separation. Given that I've now documented them here for future laughs, I'll happily go along with the consensus view.

sit23:
Happy to meet face-to-face to discuss. 

I agree that there's definitely a danger with this sort of separation that you lose track of what default file you used, etc. One suggestion to help overcome that would be to make sure the `input.nml` file is always kept in the data directory, even in `light` mode, the latter of which isn't currently the case. That way you could always reproduce the input nml that you actually ran with. `input.nml` doesn't take up much disk space, so still probably qualifies as `light`. 

I definitely take your point as well about the defaults not being appropriate in different contexts. As a suggestion for that, perhaps a good thing to do would be to have two different `namelist_baseline` files in the test cases, one for RRTM runs, and one for grey-rad runs. (Obviously we don't want too many separate files, but 2 seems OK). Even if that's not really going to change anything, it might help demonstrate that people should consider using different defaults in different contexts?

RuthG:
Looking at e.g. giant_planet_test_case.py on the diff page, even for a rather different set up a lot of parameters are shared. We could meet to discuss putting the things that really don't change straight into the model? The separate namelist base is how I'll be running the code, but obviously it doesn't have to be in the repo, I just think it makes it a bit easier for newbies to see what's going on in each file! Saving input.nml sound like a very good idea too :)

gkvallis:
Yes, having a meeting would be useful, and it is probably best as a subgroup rather than in a regular Wed meeting.
Today is out, but tomorrow we could meet late morning (.e.g. 11.30) or over lunch.



_________

Prof. Geoffrey Vallis
College of Engineering, Mathematics and Physical Sciences,
University of Exeter
http://empslocal.ex.ac.uk/people/staff/gv219/

On 23 Nov 2017, at 8:55 am, RuthG <notifications@github.com<mailto:notifications@github.com>> wrote:


Looking at e.g. giant_planet_test_case.py on the diff page, even for a rather different set up a lot of parameters are shared. We could meet to discuss putting the things that really don't change straight into the model? The separate namelist base is how I'll be running the code, but obviously it doesn't have to be in the repo, I just think it makes it a bit easier for newbies to see what's going on in each file! Saving input.nml sound like a very good idea too :)

RuthG:
As agreed yesterday, I've changed the external file from a python file to a f90 nml file. I've added a warning at the top of the nml file to encourage users to check the values set are right for their purposes. I've tested all the test case run, but have not done the bitwise comparison until we agree this is the set up we'll go with.

RuthG:
Having tested this, I'm now flip-flopping. I ran 5 tests to start: frierson, bucket, realistic continents with q-fluxes, held suarez, and axisymmetric. The first 3 of these come out with identical results, the latter 2 do not. As the only difference is what is defined in the namelists it must relate to this. 

I now see the point that, when a base namelist file is used, many parameters are set that are not necessary for the run, but which may nonetheless affect it. This makes it hard to correct the differences, but more importantly makes the input.nml files saved a bit misleading as a user, as many of the parameters included didn't really need setting in the first place.

My current feeling is to leave these test cases the way they are. I imagine if people want to streamline their experiments they'll be capable of creating their own namelist or python input files. I could turn this pull request into an example of how to do this if people think this would be useful. I'm of course also happy to continue as planned and to tidy up these differences if that's still the preferred solution, I just now see some more of the problems with it!

PR #20, opened 2017-11-21
Added 'radius' as argument in function call, L14

mp586:
Tiny change to scripts/cell_area.py - argument missing in function call


sit23:
Well spotted @mp586! Change looks good to me.

PR #21, opened 2017-11-24
Bucket depth ii

RuthG:
I realised there are some minor discrepancies between the GMD paper outline of the bucket scheme and what is coded. This pull request addresses these by:
- Changing the max bucket depth over land to have a default value of 15cm
- If the bucket is more than 75% full it now evaporates like the ocean, as described in the paper

Additionally, the initial bucket depth for ocean has been increased (though this has no effect unless the ocean completely boils off!)


sit23:
Looks good. I didn't know nested `where` statements like that were a thing.

RuthG:
I've now used the test cases to check this looks ok vs the master. I ran bucket_hydrology, frierson, realistic_continents (fixed SST), and top_down_test. The output is the same between the master and this branch, except for bucket_hydrology, which is to be expected, as the bucket formulation has changed.

Let me know if there are any other tests anyone would like to see done, and otherwise I'll merge this tomorrow.

RuthG:
I will delete this pull request and move to using the one I just created, unless anyone thinks this is a bad idea.

PR #24, opened 2017-11-29
Automated testing procedure

sit23:
Problem this p/r aims to solve is that testing a new p/r can be quite laborious, especially checking if your new code stops somebody else's test case from running, or stops it from producing the same answer as it did before.

The `trip_test.py` script automates this kind of comparison. It takes as input two commit IDs, being the latest commit you want to test, and an old commit you'd like to compare to (e.g. 'v1.0'). It then runs through each of the test cases and compares their outputs with the two different commits. If the outputs are identical, the test is counted as passed. If they are different, the tests are counted as failed.

The suggested process for future pull-requests is therefore that anybody who opens a p/r will run this script, and post the output as part of the p/r. If any of the tests fails, a justification should be given.


sit23:
I've now moved over to a command-line runner for the trip test, and have added a `README`. Hopefully this should make the tests easier and more transparent to run.

sit23:
Have added option to specify online repo that commits are checked out from, therefore making it easy to check commits on your fork rather than checking existing commits on `Execlim/Isca`.

RuthG:
Thanks! Looks good at a glance, I'll have a go with this tomorrow when I test the land humidity change 👍

jamesp:
Let's get this one in too!

sit23:
Cool - I'll merge it in then, and that will make it easier for Ruth to play with today too.

PR #25, opened 2017-11-29
A few compile and git changes to make code more portable

sit23:
When running on Exeter's HPC I ran into a couple of problems:

* The `git -C` command used in `codebase.py` is only in git versions more recent than 1.8. So Exeter's HPC (which was running 1.8) failed on these commands.
* The `mppnccombine.x` compile script still wanted to load a emps-gv specific netcdf library, rather than the appropriate env file.

To fix these problems I've:
* Added a `try` `except` statement around the git commands and have added some git commands that do the same thing in previous versions of git.
* Added the relevant `source` statement to `compile_mppnccombine.sh`. This then uses the same libraries to compile `mppnccombine.x` as it does to compile the model. However, this means that these same modules need to be loaded when running `mppnccombine.x`. I've therefore written a short shell script that loads the modules and runs `mppnccombine.x`, and have added this to the python library. This means that a normal user with the correct libraries in their computer's `env` file should have the `mppnccombine.x` compilation and execution done without them noticing.

### DIFF
--- a/postprocessing/compile_mppn.sh
+++ b/postprocessing/compile_mppn.sh
@@ -11,18 +11,18 @@ ppdir=./                 # path to directory containing the tool for combining d
 
 # 2. Load the necessary tools into the environment
 module purge
-module load netcdf-4.3.0-openmpi-intel
+source $GFDL_BASE/src/extra/env/$GFDL_ENV
 module list
 netcdf_flags=`nf-config --fflags --flibs`
 #--------------------------------------------------------------------------------------------------------
 # compile combine tool
 #cd $ppdir
-cc -O -c `nf-config --cflags` mppnccombine.c
+$CC -O -c `nf-config --cflags` mppnccombine.c
 if [ $? != 0 ]; then
     echo "ERROR: could not compile combine tool"
     exit 1
 fi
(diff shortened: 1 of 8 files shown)


jamesp:
This looks good @sit23.  A couple of points.

1. I learnt about `exec` this week.  I think you probably want to use exec in your new `mppncombine_run.sh` script.  See the change I made to `run.sh` in #26 .
More information in this stack overflow answer here: https://stackoverflow.com/a/38581329

2. It might be nice to pull that `try:... except:...` for git into a helper function rather than copy and pasting it for two separate calls.

sit23:
Thanks for suggestions @jamesp. Have made the changes you suggested. Let me know if they look alright? Both have been tested and work as before.

jamesp:
Looks great @sit23, I'm happy for this to go in.

sit23:
Just realised that I've added the `land_humidity_prefactor` change from @RuthG's P/R to this P/R by mistake.

sit23:
I think I'll just undo that change in this P/R with a new commit then will merge this in.

jamesp:
sounds good

PR #26, opened 2017-12-01
Run isca without writing a python script & interpolator update

jamesp:
This P/R includes:

* A new executable script `exp/run_isca/isca` that can be used to compile and run without needing to write a python script for each experiment.  See `exp/run_isca/README.md` for more info.
* Changes to DiagTable to support the `isca` script.  If you have a diag_table file already written to disk you want to use, wrap it in a `DiagTableFile('path/to/diag_table')` object.
* Update to `isca.util.interpolate_output` function so that it can interpolate onto it's own original pfull levels.
* A small change to the `run.sh` script so that it now `exec`s the mpirun task.  This should hopefully catch the rest of the edge cases of when you hit Ctrl-C on a run but it doesn't kill all the mpi processes.


RuthG:
Is the command line running the same as in the other pull request? I've not looked through that part in detail but the changes to the interpolator part look good. With that addition, the "MODEL" option could probably be eliminated, as this is what I put it there to do, I just added it only for the 40 level set-up I was using at the time.

jamesp:
Yep, I forgot that I had already put a PR in for the command line script!  We can get rid of the "MODEL" option if you like, I wasn't sure where those numbers had come from.

jamesp:
Sorry about the Jenkins confusion, I was playing around with it over xmas as a possible way to automatically run @sit23's trip test.  I've renamed the Jenkinsfile for now to stop it being picked up.  As there are lots of small commits in this PR, I suggest someone merge it using the "Squash and merge" command in the drop down of the "Merge Pull Request" button.

jamesp:
We can also then close #22 as that is a subset of this lot of commits.

sit23:
Just tested this and it works fine. I'm happy to do the 'Squash and merge' command if you're happy, @jamesp?

jamesp:
talking about it with @gregcolyer , it might be better just to stick to a regular merge - we might end up in a pickle with branches out of sync otherwise

sit23:
OK - I'll merge it now

PR #27, opened 2017-12-01
Resolutions script

gregcolyer:
Minor change to resolutions.py to show overhead differently (which you may or may not prefer!)
Needs testing with py3


gregcolyer:
This can be merged as soon as tested using Python 3

jamesp:
Tested with py3, looks good

PR #28, opened 2017-12-03
Missing namelist-writes to logfile

gregcolyer:
This pull request is for adding missing namelist-writes to the logfile. Here is my best attempt at the list (which I've checked as far as it goes, but may be incomplete):

badType1_nml
badType2_nml
constants_nml
horiz_interp_spherical_nml
interpolator_nml
missingVar_nml
qflux_nml
test_axis_utils_nml
test_horiz_interp_nml
test_nml

Some of these may not matter. The one I was already aware of, constants_nml, is fixed by 
f48f4a4, which has been tested (once), and may serve as an example.

This pull request could be left open until they are all done (or all the ones that matter), or closed earlier for convenience. Even without closing the pull request, the branch could be merged into master at any time.

### DIFF
--- a/src/shared/constants/constants.F90
+++ b/src/shared/constants/constants.F90
@@ -36,7 +36,9 @@ module constants_mod
 !   Constants are accessed through the "use" statement.
 ! </DESCRIPTION>
 
-use fms_mod,               only: open_file, check_nml_error, mpp_pe, close_file, write_version_number, error_mesg, NOTE
+use fms_mod,               only: open_file, check_nml_error, &
+                                 mpp_pe, mpp_root_pe, stdlog, &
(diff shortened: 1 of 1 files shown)


gregcolyer:
Previous comment in this pull request states: "Even without closing the pull request, the branch could be merged into master at any time." Now know (from pull request #35) that this is incorrect. If the branch itself is merged, which would include all outstanding commits, then the pull request is closed automatically. If not completely merged (e.g. up to some commit that is not the last commit), it would presumably stay open.

gregcolyer:
So we may need to close this pull request, if we want to merge the first (and if it remains the only) commit

gregcolyer:
Closed for pending release; reopen later for any additional namelists (cf. list above)

PR #29, opened 2017-12-06
Modify surface_flux.F90 so that land_humidity_prefactor can be set in…

RuthG:
… or outside the bracket in the evaporation equation.

Using a land humidity prefactor outside the brackets appears to allow a larger range of other land parameters to be run stably, e.g. shallower mixed layer depths, by avoiding instabilities that appear to come from downward latent heat fluxes ('condensation').

This pull request has been based around release-v1 so the release can be fast forwarded for consistency with the paper if we want. I've run this version for a couple of months to check it's stable. I'm now going to update the script on my workbranch and run overnight as a more comprehensive test. I'll do some bitwise comparisons tomorrow too.

### DIFF
--- a/src/coupler/surface_flux.F90
+++ b/src/coupler/surface_flux.F90
@@ -262,7 +262,7 @@ logical :: raoult_sat_vap        = .false.
 logical :: do_simple             = .false.
 
 real    :: land_humidity_prefactor  =  1.0    !s Default is that land makes no difference to evaporative fluxes
-real    :: evap_prefactor  =  1.0    !s Default is that land makes no difference to evaporative fluxes
+real    :: land_evap_prefactor  =  1.0    !s Default is that land makes no difference to evaporative fluxes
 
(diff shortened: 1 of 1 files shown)


sit23:
Looks great Ruth - thanks! 👍  Only thing is that whilst changing `land_humidity_prefactor` -> `land_humidity_factor_b` is sensible from a nomenclature point of view, it will mean that all of our scripts that use the prefactor will need updating to keep the same functionality. Could I tentatively suggest we keep the name `land_humidity_prefactor` as it is, and keep your new addition as `land_humidity_prefactor_a`? I know that's less satisfactory, but it would make the update backwards-compatible.

RuthG:
Changed to call it evap_prefactor to distinguish the too... hopefully this makes some sense, one prefactor is modifying the land humidity, one the evaporation. Open to suggestions if anyone can think of something clearer though!

sit23:
I like that. Evap-prefactor seems sensible to me. Thanks for doing that! :)

RuthG:
This is a minor change and the tests all came back ok, so I'll get this merged in.

PR #30, opened 2017-12-08
Qflux updates and plev.x compile script

sit23:
A few updates to the `caluclate-qflux` routines:

* Fixed some `import` statements that weren't finding the routines in the directory above.

* Added some functionality to cope with `flux_sw` and `flux_lw` arrays that are 3D (as is the case for grey radiation). Previous functionality only allowed for 2d arrays, as are output when using RRTM. This means that qfluxes can be calculated for runs with grey radiation.

Have also added a compile script for the pressure-level interpolator `plev.x`, and have updated the README accordingly.

### DIFF
--- a/postprocessing/plevel_interpolation/README
+++ b/postprocessing/plevel_interpolation/README
@@ -1,32 +1,6 @@
 Exeter instructions
 
-This set of scripts is able to take 3D data that is outputed on model sigma levels, and interpolate it to be on pressure levels. Basic instructions are as follows:
-
-cd postprocessing/plevel_interpolation/exec/
-
-source ../../../src/extra/loadmodule
-
-../bin/mkmf -p plev.x -t ../bin/mkmf.template.ia64 -c "-Duse_netCDF" -a ../src ../src/path_names ../src/shared/mpp/include ../src/shared/include
-
-make -f Makefile
-
-Once this is done the executable `plev.x` will be created in the exec folder.
-
-To execute the script, do
-
-cd ../scripts
-./plevel.sh INSERT OPTIONS HERE
-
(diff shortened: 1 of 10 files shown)


sit23:
I've tested each of these scripts again, and am happy they all work. Unless any objections, I will merge this in tomorrow.

sit23:
Also, @mp586 I made your suggested change in `calculate_qflux.py` regarding the `topo_present` variable. Let me know if it looks alright here?

mp586:
Looks good to me, only a cosmetic suggestion would be to change 'topo_present' to 'use_interpolated_.....' in the nc_file_io_xarray as well for consistency but that won't affect anything really.

The other thing that I found was to get it to work, I had to change the var_name entry in line 27 of calculate_qflux.py to 'ocean_qflux' instead of output_file_name, but that might just be related to my editing of the file paths etc. 

Sorry again that I didn't make a pull request with my changes earlier to save you the trouble :(

sit23:
Thanks for suggestions @mp586. I've made the `nc_file_io_xarray` changes you suggested. I haven't made the second change you suggested, just because the fortran assumes the q-flux file name and the field-name are the same, so I'd like to enforce that to be true in the python script too, if that's alright? I've added some comments to the script and the fortran to hopefully make that clearer. Would like to see the error you had sometime though, as it would be good to check.

I have tested all of the above code though, so I think it's ready to merge.

mp586:
Ok I didn't know that it probably is related to the way I structure my experiments and folders which is a bit different so the file paths didn't work for me, it would create the q flux file but the actual variable was missing but it worked when I put 'var_name':'ocean_qflux' 
I'll try it again and get back to you

PR #31, opened 2017-12-08
Fix for when running from an old commit that does not have a mppncombine_script

jamesp:
- Experiment.run now checks if the mpppncombine_run.sh script exists
  in the checked out code.  If it doesn't, use the GFDL_BASE version


sit23:
Looks good to me - thanks for fixing that, @jamesp. And thanks for testing it @mp586.

PR #32, opened 2017-12-14
New namelist parameters for Held-Suarez forcing and spectral dynamics

gregcolyer:
Adds `P00` (reference pressure used in Held-Suarez, not necessarily the same as surface pressure) to the namelist, and a new parameter `lapse_rate` for [adiabatic] lapse rate normalized to dry. `lapse_rate` is used only for classic newtonian_damping (original and exoplanet variants); there is also a similar but dimensional parameter `lapse` used only for top_down_newtonian_damping; the two could be consolidated (not done).

Also adds a new parameter `reference_temp` to the spectral dynamics namelist, previously hard-coded to 300 K.


sit23:
Looks fine to me, but happy to defer this to @Alexander-P and @jamesp as I don't ever use the Held-Suarez bit of the code.

jamesp:
Looks good.  We should probably get rid of the `alpha` parameter that I added (and haven't really used) as this achieves the same thing

gregcolyer:
Ah yes, I didn't notice `alpha`. Same as `(KAPPA*lapse_rate)`?

jamesp:
Yep.  In fact, let's just scrap the EXOPLANET2 conditional branch altogether

Alexander-P:
I didn't notice that there was an existing normalized lapse rate thing, I may look into using that for top_down. This looks good to me

gregcolyer:
There wasn't until now

jamesp:
thanks @gregcolyer, I was just doing this!

Alexander-P:
good reading comprehension, me!

gregcolyer:
I think the default `lapse` (in K/km?) corresponds to a `lapse_rate` of 2/3, whereas 1 is needed for standard Held-Suarez. We could code different defaults for classic and top_down though, if you want to keep top_down the same.

gregcolyer:
Following discussion with @Alexander-P: no need for different defaults, so I will code up this consolidation (remove `lapse` from namelist but set based on `lapse_rate`) and push here

gregcolyer:
Side note: `P00` defaults to 1E5 exactly, as in the Held-Suarez paper; this does not have to equal the reference surface pressure, but for consistency one should presumably set the latter to 1E5 (i.e. override the more precise default value) in the typical case. It's probably not appropriate to change either default (one should match the paper, the other should be the best available value for Earth), but perhaps the code should warn if they differ only by this order (i.e. a few per cent)?

jamesp:
Shall we put this in, @gregcolyer?

jamesp:
shall we put this in, @gregcolyer ?

gregcolyer:
Testing now

PR #33, opened 2017-12-14
Python 2.7 & 3.x compatability

jamesp:
Small fixes to the python framework to be compatible with python 2.7 and 3.x.


sit23:
Just tested this and it works perfectly with both python 2.7 and 3.6, so very happy for it to be merged in.

PR #34, opened 2017-12-15
Only grey sigma fix

sit23:
PR #1 was put in to fix a normalisation problem for lw optical depths in BOG and Geen grey radiation schemes. But I also added some changes to the way surface pressure was defined, which proved more controversial, so it wasn't merged. This PR is just the normalisation, so should be straight forward to approve.


sit23:
Running the trip-test on this now. Will report back when they're done.

RuthG:
Code looks good to me! Happy for you to merge it once the trip test is done.

sit23:
```Results for all of the test cases ran comparing 6faf53d and 5b01c2c are as follows...
variable_co2_grey : fail
held_suarez : pass
variable_co2_rrtm : pass
bucket_model : pass
top_down_test : pass
frierson : pass
axisymmetric : pass
MiMA : pass
giant_planet : pass```
All tests came back fine, apart from the one that uses BOG, which fails (I think) because of the change to the long-wave optical depth normalisation, so it's an expected fail. I think this is therefore alright to put into the master.

PR #38, opened 2018-01-16
External namelist test

RuthG:
Having attempted to test the external_namelist pull request, I now agree that it is somewhat confusing to have a base namelist which changes parameters that are not relevant for all configurations.

I suggest we instead provide an example of how namelist options can be read in for a case where we currently have 2 scripts doing almost the same experiment. I've chosen the realistic_continents case, and moved the shared parameters to a common fortran namelist to be read in, and kept the differences between the two in the python scripts. I have tested that the results are the same as when the current master is used. 

Thoughts welcome! - I'm still happy to go back through the previous pull request and locate the differences, but I now see the advantages in having python scripts which include only relevant parameter changes. Apologies for the late change of heart!

If no-one dislikes this way, I'll merge this tomorrow and delete the previous pull request.

### DIFF
--- a/exp/test_cases/realistic_continents/namelist_basefile.nml
+++ b/exp/test_cases/realistic_continents/namelist_basefile.nml
@@ -1,5 +1,6 @@
-! For sets of experiments which will use predominantly the same parameters, it may be tidier to read these from an external namelist file. This example illustrates how to do this for two experiments configured with realistic continents, one using fixed SSTs, one using Q-fluxes.
-
+! For parameters which are frequently set, in order to keep python scripts tidy and clear, it may be useful to read these from an external namelist file.
+! This example illustrates how to do this for two experiments configured with realistic continents, where the only differences are the choice of surface condition, with one using fixed SSTs and the other Q-fluxes.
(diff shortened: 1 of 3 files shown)


gregcolyer:
Looks good to me. `do_qflux` is the same for both, so could be moved to the basefile (or even omitted, since false is the code default anyway), but I suppose it should be written however is logically the clearest. Does the basefile make sense on its own, i.e. would it be a meaningful case without any overrides?

sit23:
Just been trying to run this, and the script is failing on me on the line where the namelist is read ` nml = f90nml.read(namelist_name)`. The error message is just `StopIteration`. I've tried it with versions 0.21 and 0.23 of f90nml and neither seems to work. :( Presumably something is different in your setup @RuthG - can we compare tomorrow sometime?

sit23:
Also, before it's merged in, I have a feeling (though I can't test it atm b/c of the above) that this change will stop the trip test from working. The trip-test relies on importing the `namelist` variable from each of the test experiments, which has been removed here. Once I can run the above I'll check this out.  I hadn't made it clear that this was how the trip-tests worked, so no worries about this @RuthG. Once I can run the above I'll make some changes to the trip-test so that it will work as before.

sit23:
This works now for me with f90nml version 0.23, so very happy for it to go in. I've re-written the trip-test function to import the `exp` object rather than the `namelist` variable, so there should be no problems there either. I'll make a separate P/R for that shortly.

PR #39, opened 2018-01-17
Update to trip test

sit23:
Updated the way the trip test imports namelists so that test cases don't have to define the `namelist` variable in order to work.


sit23:
Running this trip test now, but after that it should be fine to go in.

sit23:
Have now updated test cases to run with py3, as py3 didn't like `current_data = [001,...]` as it said `001` was an invalid token. Runs identically when using `current_data = [1,...]` instead so have made that change. Trip test runs fine with both py2 and py3 now.

sit23:
Have tested this thoroughly now with py3 and py2, so am happy for it to be merged in. If nobody has any objections, I'll do it tomorrow.

PR #42, opened 2018-01-22
Util fix

RuthG:
Little tweak to util.py to sort out issues with specifying fields to interpolate. (Thanks jp!)


jamesp:
I've updated to remove the MODEL option and added a ValueError exception if you enter a string that is not 'input' or 'even'.  Tested on my instance

PR #44, opened 2018-03-07
Bug fix in idealized_moist_phys

jamesp:
I think this ampersand is a (possibly silent) bug in moist_phys.  ifort seems to skip over it, but gfortran fails to compile.

**I HAVE NOT YET VERIFIED BITWISE RESULTS AFTER REMOVING THIS**

### DIFF
--- a/src/atmos_spectral/driver/solo/idealized_moist_phys.F90
+++ b/src/atmos_spectral/driver/solo/idealized_moist_phys.F90
@@ -254,7 +254,7 @@ integer ::           &
      id_cape,        &
      id_cin
 
-integer, allocatable, dimension(:,:) :: & convflag ! indicates which qe convection subroutines are used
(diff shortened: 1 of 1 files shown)


sit23:
Will run the trip test on this and check :)

sit23:
```Results for all of the test cases ran comparing 5c4415b and 5868cfd are as follows...
axisymmetric : pass
bucket_model : pass
frierson : pass
giant_planet : pass
held_suarez : pass
MiMA : pass
top_down_test : pass
variable_co2_grey : pass
variable_co2_rrtm : pass```
This compares your deleted `&` version `5c4415b` with the current state of the master `5868cfd`, and all the test cases produce identical output.

sit23:
Happy to merge this in, given that the answers are the same.

jamesp:
Great, thanks @sit23.  Getting rid of that ampersand is necessary to compile on gfortran

PR #45, opened 2018-03-14
Isca running on gfortran

jamesp:
I won't commit this to master just yet as there are a few changes to the python framework that need a bit more testing, but it's pretty close if others want to play with it.  This PR is to put the changes on the `docker` branch of ExeClim from my fork.

Changes:
* Added a compile template for using gfortran instead of ifort.
* Added a Dockerfile for building an image that contains the isca executable and necessary mpi and netcdf libraries to run on multiple cores.
* The `isca.util.run_cli` function that gives an easy way to give a python experiment script a command line interface.
* T21 option in the predefined resolutions in `experiment.py`

Bug fixes:
* `atmos_model.F90:384` did not write the number of minutes to the restart file correctly for NO_CALENDAR and gfort, which seems to default to 1 where ifort default was 0.  This resulted in a 60second drift per run break.
* `idealized_moist_phys.F90:257` has an inline `&` that `gfortran` errors on.  `ifort` appears to be unaffected, but is potentially doing something undefined?

The  Dockerfile has been used to create image https://hub.docker.com/r/jamespenn/isca/ on Docker Hub.  This should make it easy to run Isca on any machine, albeit with a potential performance hit.  See link above for details on how to use the container.


Dockerfile
```diff
@@ -0,0 +1,36 @@
+FROM ubuntu:latest
+
+ENV GFDL_WORK /tmp
+ENV GFDL_BASE /isca
+ENV GFDL_DATA /data
+ENV GFDL_ENV docker
+
+# ignore missing hardware needed lfor openMPI speedup
+ENV ["OMPI_MCA_btl", "^openib"]
```
jamesp: This is only necessary on the mac host, so it can probably be removed

sit23:
This looks great - thanks @jamesp. Is there anything you want to do before putting this p/r in?

jamesp:
The docker bits of this seem pretty stable now.  Over the weekend I ran isca on isca, compiled with gfortran.  I've now got an ifort equivalent run in the parallel queue.  Once that's run and I've confirmed some basic sanity checks I think it'll be ready to merge

sit23:
Perfect - sounds great. One other quick question - what method are you using to tell the front-end to compile using the gfortran mkmf template? I can't see where you've specified it in the above. Could it be an option to the `cb.compile()` command, or something like that?

jamesp:
it's currently a new env var: `export GFDL_COMPILER=gfortran` but this isn't ideal.  I'd welcome other ideas.

jamesp:
see the docker and isca_gfortran env files for examples.  you should be able to just take that isca_gfortran and use it on isca (I hope!).  I should think it's small changes only to get it running on emps-gv

sit23:
Ah, I see. I didn't notice your changes to `compile.sh` as they are already on the `docker` branch of `Execlim/Isca`, so they didn't show up in the diff above. I think the way you're doing it seems sensible. My only suggestion would be to make the variable name something clearer - e.g. `MKMF_TEMPLATE_SUFFIX` rather than `GFDL_COMPILER`. So then it would be clear that this variable is `ia64` by default (which is kind of what you're doing in `compile.sh` at the moment), and that you could use a different mkmf template if you wanted by changing that variable name?

jamesp:
yes, that makes sense - I'll make that change

src/extra/python/isca/templates/compile.sh
```diff
@@ -15,12 +19,8 @@ template_debug={{ template_dir }}/mkmf.template.debug
 execdir={{ execdir }}        # where code is compiled and executable is created
 executable={{ executable_name }}
 
-netcdf_flags=`nc-config --fflags --flibs`
+netcdf_flags=`nf-config --fflags --flibs`
```
jamesp: Given that this seems to vary wildly between machines, it might be worth pulling it out into the environment file

src/extra/python/isca/templates/mkmf.template.gfort
```diff
@@ -0,0 +1,27 @@
+# template for the  gfortran compiler
+# typical use with mkmf
+# mkmf -t template.ifc -c"-Duse_libMPI -Duse_netCDF" path_names /usr/local/include
+CPPFLAGS = -I/usr/local/include
+NETCDF_LIBS = `nf-config --fflags  --flibs`
```
jamesp: Again here

jamesp:
I think this is ready to go.  From trying to compile on different architectures I've learnt that the `nf-config` and `nc-config` command line options are not nearly as ubiquitous as I thought.  Therefore I suggest in the future we move the netcdf library discovery into the env file, and out of the compile files so that it can be system specific. I won't do that in this PR though. @sit23 are you happy for this to merge into master?

sit23:
This looks great - happy for you to merge this @jamesp. Nice work.

PR #56, opened 2018-07-19
Bug fix in two_stream

jamesp:
array `lw_dtrans_win` is only used in the B_GEEN scheme, and is only allocated when `lw_scheme == B_GEEN` (line 278).  Therefore at line 517 `lw_dtrans_win` is potentially unallocated.  Moved this assignment into the B_GEEN scheme guard at line 528.

All other changes are whitespace.

### DIFF
--- a/src/atmos_param/two_stream_gray_rad/two_stream_gray_rad.F90
+++ b/src/atmos_param/two_stream_gray_rad/two_stream_gray_rad.F90
@@ -107,7 +107,7 @@ real    :: single_albedo      = 0.8
 real    :: back_scatter       = 0.398
 real    :: lw_tau_0_gp        = 80.0
 real    :: sw_tau_0_gp        = 3.0
-real    :: lw_tau_exponent_gp = 2.0 
+real    :: lw_tau_exponent_gp = 2.0
 real    :: sw_tau_exponent_gp = 1.0
 real    :: diabatic_acce = 1.0
 real,save :: gp_albedo, Ga_asym, g_asym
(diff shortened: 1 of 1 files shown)


src/atmos_param/two_stream_gray_rad/two_stream_gray_rad.F90
```diff
@@ -526,6 +525,7 @@ subroutine two_stream_gray_rad_down (is, js, Time_diag, lat, lon, p_half, t,
 case(B_GEEN)
   ! split LW in 2 bands: water-vapour window and remaining = non-window
   ! ref: Ruth Geen etal, GRL 2016 (supp. information).
+  lw_dtrans_win = 1.
```
jamesp: if the `lw_dtrans_win` variable is going to be used by other radiation schemes, the alternative fix to this would be to move line 278 out of the `case(B_GEEN)` guard

sit23:
With the code as it is at the moment, all the tests pass, apart from the realistic_continent_qflux one, which is not in the list below:
```
Test passed for variable_co2_rrtm. Commit f4dd14 gives the same answer as commit 77a3d49
Results for all of the test cases ran comparing 77a3d49 and f4dd14 are as follows...
axisymmetric : pass
bucket_model : pass
frierson : pass
giant_planet : pass
held_suarez : pass
MiMA : pass
top_down_test : pass
variable_co2_grey : pass
variable_co2_rrtm : pass
```

sit23:
Shall we merge this now?

jamesp:
yes, I think we should

PR #63, opened 2018-08-07
Bug fix from Paul O'Gorman, received by Geoff on August 3rd 2018.

sit23:
Paul's explanation is below:
******
I found a problem with the old boundary-layer diffusivity module that
is used in the  FMS "gray-radiation" code.

The issue is that the diffusivity can be set to non-zero spurious
values above the boundary layer when the boundary-layer height varies
a lot over a processor domain. Fortunately this doesn't seem to
strongly affect the climate in the kinds of simulations I have run
with the GCM. But a knock-on effect is that simulation output can
depend on the processor decomposition used.
******

### DIFF
--- a/src/atmos_param/diffusivity/diffusivity.F90
+++ b/src/atmos_param/diffusivity/diffusivity.F90
@@ -145,7 +145,7 @@ real    :: ampns_max           = 1.0E20  ! limit to reduction factor
 logical :: do_entrain          =.true.
 logical :: do_simple           =.false.
 
-logical :: use_pog_bug_fix     =.false.
+logical :: use_pog_bug_fix     =.true.
 
 namelist /diffusivity_nml/ fixed_depth, depth_0, frac_inner,&
                            rich_crit_pbl, entr_ratio, parcel_buoy,&
--- a/src/atmos_param/diffusivity/diffusivity.F90
+++ b/src/atmos_param/diffusivity/diffusivity.F90
@@ -145,12 +145,14 @@ real    :: ampns_max           = 1.0E20  ! limit to reduction factor
 logical :: do_entrain          =.true.
 logical :: do_simple           =.false.
 
+logical :: use_pog_bug_fix     =.true.
+
 namelist /diffusivity_nml/ fixed_depth, depth_0, frac_inner,&
                            rich_crit_pbl, entr_ratio, parcel_buoy,&
                            znom, free_atm_diff, free_atm_skyhi_diff,&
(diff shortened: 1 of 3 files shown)


sit23:
I will run the trip test on this to check its impact.

sit23:
Clearly this does make a difference to the results of the moist model:
```
Results for all of the test cases ran comparing 86893cc and a865978 are as follows...
axisymmetric : fail
bucket_model : fail
frierson : fail
giant_planet : fail
held_suarez : pass
MiMA : fail
realistic_continents_fixed_sst : fail
realistic_continents_variable_qflux : fail
top_down_test : pass
variable_co2_grey : fail
variable_co2_rrtm : fail
```

sit23:
I'll do a longer experiment to compare the climates in the Frierson model.

sit23:
I've tested the impact of the bug-fix in a 20-year run of the Frierson model, and the climate doesn't change much, if at all. So I think this change is fine to go in. This change will mean, however, that results using the moist-model will not be bit-wise identical before and after this commit. I have, however, added a namelist variable in the `diffusivity_nml` called `use_pog_bug_fix`, so that if the original behaviour, including this bug, is desired, then this can be set to false. If you set it to false, all the trip tests pass, as expected:
```
Results for all of the test cases ran comparing 86893cc and d13f6b7 are as follows...
axisymmetric : pass
bucket_model : pass
frierson : pass
giant_planet : pass
held_suarez : pass
MiMA : pass
top_down_test : pass
variable_co2_grey : pass
variable_co2_rrtm : pass
```

src/atmos_param/diffusivity/diffusivity.F90
```diff
@@ -496,6 +498,18 @@ subroutine diffusivity_pbl(t, u, v, z_half, h, u_star, b_star, &
     k_m(:,:,k) = k_m_ref*factor
     k_t(:,:,k) = k_t_ref*factor
   end where
+
+
+  if(use_pog_bug_fix) then
```
sit23: I have added the namelist option to turn the fix on and off, so that old-behaviour can be retained if desired.

src/extra/python/scripts/calendar_calc.py
```diff
@@ -1,4 +1,4 @@
-from netcdftime import utime
+from cftime import utime
```
sit23: I've also made a small update to `calendar_calc` to reflect the change of the `netcdftime` module to be `cftime`.

sit23:
As I've had no objections on this, I'm going to merge it now.

Issue #66, opened 2018-08-10
Equinox_day inconsistency


ntlewis:
`equinox_day` (the fraction of the year at which autumn equinox occurs) is set as default to `0.0` in `two_stream_gray_nml` but `0.75` in `rrtm_radiation_nml` and `socrates_nml`. 

Should the value in `two_stream_gray_nml` be set to the correct value of `0.75`?

sit23:
That was done because most people using the grey-radiation code at the time we added it weren't using the Earth calendar, so having equinox as day number 1 was useful. Whereas the typical RRTM user was using the Earth calendar, for which 0.75 is appropriate. We could change the grey-rad one though, given that more people are using that for Earth now.

PR #67, opened 2018-08-10
Change default 'equinox_day' value in two_stream_gray 

ntlewis:
Change default 'equinox_day' value in two_stream_gray from 0.0 to 0.75 inline with Socrates and RRTM.  Addresses Issue #66 .


sit23:
This was something that confused the summer school students too, so I think it would be helpful to merge this. People who want the old behaviour can very easily change the value in a namelist.

PR #68, opened 2018-08-10
Add uz, vz, wz

RuthG:
Slight modification to spectral_dynamics.F90 to allow the fluxes of geopotential height to be output, so that for example moist static energy transport can be evaluated without the need to save daily data.

All the trip tests passed, so it should be ready to merge unless anyone can see any problems.


ntlewis:
Not a code issue, bit how does this allow someone to compute an MSE flux without daily data? Surely we would also need uT and uq as outputs also? Because we are interested in vh = v(cpT + gz + Lq)?

RuthG:
You're correct, but those are already available as vcomp_temp and sphum_v. I'd just not added geopotential height * v previously as I was just looking at temperature and humidity transports.

ntlewis:
Learn something new everyday.... :)

RuthG:
😋

RuthG:
Sorry I can change the details of the request if that's better, it could come across a little confusing in the history down the line

sit23:
Looks great to me.

Issue #71, opened 2018-09-04
Run Isca - AttributeError: 'DiagTableFile' object has no attribute 'files'


AlexAudette:
Hi, I am trying to run the test1 case (default run in the run_isca directory as described in the README.md file there) and the model runs through the first month without any error message being written. However, when it comes time to write the output in the output files, I get an error message that I never got before. Here is what it says (I removed the normal lines that are not useful here) :

2018-08-31 15:02:07,751 - isca - INFO - Emptied run directory '/scratch/aaudette/gfdl_work/experiment/test1/run'
[...]
2018-08-31 15:03:48,637 - isca - INFO - Run 1 complete
Traceback (most recent call last):
  File "./isca", line 51, in <module>
    exp.run(1, use_restart=False, num_cores=args.num_cores)
  File "/home/aaudette/Isca/src/extra/python/isca/helpers.py", line 22, in _destructive
    return fn(*args, **kwargs)
  File "/home/aaudette/Isca/src/extra/python/isca/helpers.py", line 38, in _useworkdir
    return fn(*args, **kwargs)
  File "/home/aaudette/Isca/src/extra/python/isca/experiment.py", line 293, in run
    for file in self.diag_table.files:
AttributeError: 'DiagTableFile' object has no attribute 'files'

Edit: I was able to run correctly all the test cases I tried before.

Thank you for your help.
Alex

jamesp:
Hi Alex,
This bit of code is not often used, and poorly tested -- sorry about that!  you should be able to get the python files in `exp/test_cases` to run, this is the most common way of running experiments.
I'll see if I can put together a quick parser to fix that bug though, thanks for reporting it.

jamesp:
@AlexAudette I've made a fix and pushed to the master branch, you should be able to get it by pulling ExeClim/Isca:master.  Let me know if it doesn't work for you.

PR #72, opened 2018-09-04
Basic DiagTable Parser

jamesp:
Added a basic DiagTable parser to support using existing diag tables eg.
```
diag = DiagTable.from_file(‘/path/to/existing/diag_table’)
exp.diag_table = diag
```
This is to address #71


jamesp:
This adds a new method and doesn't change existing functionality.  I'm going to merge.

PR #75, opened 2018-09-11
Restructuring how Socrates code is integrated, and bug fixing

sit23:
# Fortran Changes

As mentioned in #61 , I found a disagreement in results between 6c4289d and 3d1030e. Some of this was down to #63's changes to the diffusivity mod. But I traced the other changes to d3f9678, which changed the way that `socrates_interface` outputted fluxes. It turns out the older version was more correct, and d3f9678 introduced a bug. The problem created by d3f9678 was that `socrates_calc` spat out fluxes on half levels, but I was capturing this output with 1 fewer levels than required. The upshot of this was that versions more from d3f9678 and newer have been passing out the surface fluxes from 1 level up in the atmosphere, rather than from the surface. This has lead to lower atmospheric temperatures and lower near-surface specific humidity values. The changes included in this P/R fix these problems. 

# Socrates Codebase changes:

In order to complete the above testing, I wanted to make the socrates test case work as part of the trip tests. To do this, I created a new environment variable called `GFDL_SOC`, which points to the location of the socrates source code (meaning this can be kept outside of the Isca folder on a local machine). Upon creation of the `SocratesCodebase` object, a sym-link is created from the location of the Socrates source to `Isca/src/atmos_param/socrates/src/trunk`, which is the location expected for the Socrates source code. This allows the socrates test case to work with the trip test. I will update the Socrates Readme accordingly.

### DIFF
--- a/exp/test_cases/socrates_test/socrates_aquaplanet.py
+++ b/exp/test_cases/socrates_test/socrates_aquaplanet.py
@@ -84,6 +84,8 @@ exp.namelist = namelist = Namelist({
         'dt_rad':3600,
         'store_intermediate_rad':True,
         'chunk_size': 16,
+        'use_pressure_interp_for_half_levels':False,
+        'tidally_locked':False
     }, 
     'idealized_moist_phys_nml': {
         'do_damping': True,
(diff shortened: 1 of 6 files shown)


ntlewis:
@sit23 If `GFDL_SOC` isn't set, will the model default to `Isca/src/atmos_param/socrates/src/trunk` or just not find the socrates code?

sit23:
@ntlewis - if `GFDL_SOC` isn't set then it will default to looking in the folder you mentioned, so it compile as before without current users having to make any changes.

PR #84, opened 2019-03-01
Two stream scatter

ntlewis:
Code that solves the two-stream radiative transfer equations with scattering in Isca (semi-grey; i.e., grey in shortwave and longwave). 

Follows Pierrehumbert, 2010: Principles of Planetary Climate. Chapter 5, Page 353.

New code added in directory atmos_param/two_stream_scatter_rad, along with .pdf documentation. Code in idealized_moist_phys, as well as various path_names files has been modified for coupling. These should be the only changes. 

This code is an enhancement. With the addition of clouds into Isca, this provided a grey code that can handle scattering for a finite depth, inhomogeneous atmosphere. 

This code probably needs a little more testing before being committed to main repo, but I'm submitting this P/R now in case anyone wants to have a play. I have however used the code to reproduce one of the figures in Pierrehumbert (2010), and the code does this perfectly (Figure 5.8).


ntlewis:
Hi All, this code compiles and runs as intended. I'm not sure what else needs to be done with this commit. I'll check over the p/r to see if I've included any accidental garbage files. I could add an example test case, however I think this is overkill for something that is only a small enhancement. In the future, it might be worth adding a test case that has this radiation + simple clouds, as unlike the other grey schemes this one can handle multiple scattering in the shortwave. When we write the documentation, I'll make sure this is fully documented alongside the other grey codes. @sit23 or @RuthG  would it be possible to 'review' this so that it can be committed to the main repository?

ntlewis:
> Hi All, this code compiles and runs as intended. I'm not sure what else needs to be done with this commit. I'll check over the p/r to see if I've included any accidental garbage files. I could add an example test case, however I think this is overkill for something that is only a small enhancement. In the future, it might be worth adding a test case that has this radiation + simple clouds, as unlike the other grey schemes this one can handle multiple scattering in the shortwave. When we write the documentation, I'll make sure this is fully documented alongside the other grey codes. @sit23 or @RuthG would it be possible to 'review' this so that it can be committed to the main repository?

NB: probably best to merge this after the column model, as i'll need to add it to the column model's path_names file.

PR #88, opened 2019-05-01
10m winds & 2m temperature calculations and output fields

mp586:
I've commented the parts where I changed code in surface_flux.F90 and in idealized_moist_phys.F90 with "!mp586 for 10m winds and 2m temp" 

Here's how to get the new output fields

diag.add_field('atmosphere', 'temp_2m', time_avg=True) # add 2m temperature
diag.add_field('atmosphere', 'u_10m', time_avg=True) # add 10m zonal winds
diag.add_field('atmosphere', 'v_10m', time_avg=True) # add 10m meridional winds 

The only caveat is I haven't actually looked at the output yet to see whether it's realistic, but it certainly outputs the fields! Sorry about that I'll try to look at them and report back. However I know Qun needs the 2m winds and there is no point in implementing that a second time ;)


sit23:
Nice work Marianne - this looks great. I notice that your p/r includes lots of your experiment files (.py and .nc). I think it would be worth us creating a new branch with just the fortran changes in, and then using that to make a new p/r, so that those are the only files that go into the master. I'm happy to do this if you like, or I can send you an email with some instructions?

sit23:
This P/R has been superseded by #91. I am therefore closing it.

PR #89, opened 2019-07-23
Rad flux diagnostics

penmaher:
Added three extra flux outputs. All fluxes are positive in value.

### DIFF
--- a/exp/test_cases/socrates_test/socrates_aquaplanet.py
+++ b/exp/test_cases/socrates_test/socrates_aquaplanet.py
@@ -30,36 +30,35 @@ inputfiles = [os.path.join(GFDL_BASE,'input/rrtm_input_files/ozone_1990.nc')]
 
 #Tell model how to write diagnostics
 diag = DiagTable()
-#diag.add_file('atmos_monthly', 30, 'days', time_units='days')
-diag.add_file('atmos_ts', 15, 'minutes', time_units='minutes')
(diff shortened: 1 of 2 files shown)


penmaher:
PullRequest89

Left plot shows the fluxes are only incremented once for every rad timestep (as expected) and right is to shown sign for each of the flux components.

exp/test_cases/socrates_test/socrates_aquaplanet.py
```diff
@@ -45,16 +45,22 @@
...
+diag.add_field('socrates', 'soc_surf_flux_sw_down', time_avg=True)
+#net (up) TOA and downard fluxes
 diag.add_field('socrates', 'soc_olr', time_avg=True)
-diag.add_field('socrates', 'soc_toa_sw', time_avg=True)
+diag.add_field('socrates', 'soc_toa_sw', time_avg=True) 
+diag.add_field('socrates', 'soc_toa_sw_down', time_avg=True)
+
 # additional output options commented out 
-#diag.add_field('socrates', 'soc_flux_lw', time_avg=True)
-#diag.add_field('socrates', 'soc_flux_sw', time_avg=True)
```
sit23: It would be great if these two commented lines could be left in?

penmaher: done

PR #90, opened 2019-07-31
Updating outputs in realistic continents case, and updating Socrates Readme

sit23:
I realised that we don't have all the outputs required to do vertical interpolation in the realistic-continents test case. I've therefore added zsurf, which means we have now have sufficient outputs to do this interpolation. These are: `ps, pk, bk` and `zsurf`.


sit23:
Issues have been raised in Issue #93 regarding how to compile Isca with newer versions of Socrates than the one we have been using. The problem is that we use a static `path_names` file to tell the compiler what Socrates files to compile, but if these file names are changed in a new version of Socrates, then the compilation will fail. In the future, I'd like `path_names` for socrates to be generated automatically. However, for now, I've updated the Readme for the Socrates test case with instructions on what to do if this happens.

sit23:
Following further discussion of issues #93 and #94, I have updated the Readme further to advise on errors that might arise when upgrading to newer versions of Socrates. This advice is sufficient to solve all of the problems in #93 and #94, so once this PR is merged, I will close these issues.

sit23:
Thanks @lqxyz. I'll merge this P/R now.

Issue #105, opened 2020-01-30
Plevel interpolator compile script uses old mkmf template


sit23:
Need to update `compile_plev_interp.sh`

PR #107, opened 2020-01-31
Plev compile fix

sit23:
As per issue #105, it was noted by users in Bristol that the `compile_plev_interpolation.sh` script was giving errors. I found that the reason was that it was trying to use the mkmf template from the `postprocessing/bin` directory, rather than the mkmf template used to compile the model. I had fixed this problem on the `pre_ictp_mods` branch back in August 2018, but hadn't merged it in. So apologies for that.

I have now updated the compile script so that it picks the same mkmf template as the model uses, and I've also added a simplified version of my `run_plevel.py` script, which was also taken from the `pre_ictp_mods` branch. There is no need to run the trip test in this case, as no fortran modifications have been made.

### DIFF
--- a/postprocessing/plevel_interpolation/compile_plev_interpolation.sh
+++ b/postprocessing/plevel_interpolation/compile_plev_interpolation.sh
@@ -5,6 +5,6 @@ cd ./exec
 
 source $GFDL_BASE/src/extra/env/$GFDL_ENV
 
-../bin/mkmf -p plev.x -t ../bin/mkmf.template.ia64 -c "-Duse_netCDF" -a ../src ../src/path_names ../src/shared/mpp/include ../src/shared/include
+../bin/mkmf -p plev.x -t $GFDL_BASE/src/extra/python/isca/templates/mkmf.template.ia64 -c "-Duse_netCDF" -a ../src ../src/path_names ../src/shared/mpp/include ../src/shared/include
 
 make -f Makefile
--- a/postprocessing/plevel_interpolation/compile_plev_interpolation.sh
+++ b/postprocessing/plevel_interpolation/compile_plev_interpolation.sh
@@ -5,6 +5,9 @@ cd ./exec
 
 source $GFDL_BASE/src/extra/env/$GFDL_ENV
 
-../bin/mkmf -p plev.x -t $GFDL_BASE/src/extra/python/isca/templates/mkmf.template.ia64 -c "-Duse_netCDF" -a ../src ../src/path_names ../src/shared/mpp/include ../src/shared/include
+compiler=${GFDL_MKMF_TEMPLATE:-ia64}
+template=mkmf.template.${compiler}
+
+../bin/mkmf -p plev.x -t $GFDL_BASE/src/extra/python/isca/templates/$template -c "-Duse_netCDF" -a ../src ../src/path_names ../src/shared/mpp/include ../src/shared/include
 
 make -f Makefile
--- a/postprocessing/plevel_interpolation/scripts/run_plevel_simple.py
+++ b/postprocessing/plevel_interpolation/scripts/run_plevel_simple.py
@@ -1,0 +1,51 @@
+from netCDF4 import Dataset  
+from plevel_fn import plevel_call, daily_average, join_files, two_daily_average, monthly_average
+import sys
+import os
+import time
+import pdb
+import subprocess
+
+start_time=time.time()
+base_dir='/home/sthomson/datadir_isca/'
+exp_name_list = ['project_7_ice_albedo_False_t42']
+avg_or_daily_list=['monthly']
+start_file=1
+end_file=7
+nfiles=(end_file-start_file)+1
+
+level_set='standard' #Default is the standard levels used previously. ssw_diagnostics are the ones blanca requested for MiMa validation
+mask_below_surface_set=' ' #Default is to mask values that lie below the surface pressure when interpolated. For some applications, you want to have values interpolated below ground, i.e. as if the ground wasn't there. To use this option, this value should be set to '-x '. 
+
+try:
+    out_dir
+except:
+    out_dir = base_dir
+
+plevs={}
+var_names={}
+
+if level_set=='standard':
+
+    plevs['monthly']=' -p "1000 10000 25000 50000 85000 92500"'
+    plevs['6hourly']=' -p "1000 10000 25000 50000 85000 92500"'
+    plevs['daily']  =' -p "1000 10000 25000 50000 85000 92500"'
+    
+    var_names['monthly']='-a slp height'
+    var_names['6hourly']='-a slp height'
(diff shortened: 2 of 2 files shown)


sit23:
As we discussed this yesterday, maybe @wseviour could do the code review for me?

sit23:
I've also added a change to the script created for removing restart files for data clearup.

sit23:
Would one of @RuthG, @penmaher or @wseviour mind reviewing this for me so that I can merge it in? Nothing very substantial.

postprocessing/plevel_interpolation/compile_plev_interpolation.sh
```diff
@@ -5,6 +5,9 @@ cd ./exec
 
 source $GFDL_BASE/src/extra/env/$GFDL_ENV
 
-../bin/mkmf -p plev.x -t ../bin/mkmf.template.ia64 -c "-Duse_netCDF" -a ../src ../src/path_names ../src/shared/mpp/include ../src/shared/include
+compiler=${GFDL_MKMF_TEMPLATE:-ia64}
```
penmaher, line 8: Do we want to add any other templates to the folder or is the idea people can add them if they want? Just wondering if we have other templates sitting on different branches that could come for free in the PR.

postprocessing/plevel_interpolation/scripts/run_plevel_simple.py
```diff
@@ -0,0 +1,51 @@
+from netCDF4 import Dataset  
+from plevel_fn import plevel_call, daily_average, join_files, two_daily_average, monthly_average
+import sys
+import os
+import time
+import pdb
+import subprocess
+
+start_time=time.time()
+base_dir=os.environ['GFDL_DATA']
```
penmaher, line 10: This is a nice change here. In the other example this was a hard coded link to a path name.

postprocessing/plevel_interpolation/scripts/run_plevel_simple.py
```diff
@@ -0,0 +1,51 @@
...
+            print(n+start_file)
+
+            nc_file_in = base_dir+'/'+exp_name+'/run%04d'%(n+start_file)+'/atmos_'+avg_or_daily+'.nc'
+            nc_file_out = out_dir+'/'+exp_name+'/run%04d'%(n+start_file)+'/atmos_'+avg_or_daily+file_suffix+'.nc'
+
+            if not os.path.isfile(nc_file_out):
+                plevel_call(nc_file_in,nc_file_out, var_names = var_names[avg_or_daily], p_levels = plevs[avg_or_daily], mask_below_surface_option=mask_below_surface_set)
+
+print('execution time', time.time()-start_time)
+
```
penmaher, line 51: This is much cleaner. Thanks for the example.

src/extra/python/scripts/remove_certain_restart_and_data_files.py
```diff
@@ -12,46 +13,58 @@ def __init__(self, basedir, workdir, datadir, exp_name):
...
-    datadir = os.environ['GFDL_DATA']
+    basedir = os.environ['GFDL_BASE']
     expname = '/'+exp_name+'/'
 
     exp_object = temporary_exp_object(basedir, workdir, datadir, exp_name)
     
     return exp_object
 
 
-def keep_only_certain_restart_files(exp_object, max_num_files, interval=12):
```
penmaher, line 27: Why comment this function out and not delete it? Do you want the legacy?

src/extra/python/scripts/remove_certain_restart_and_data_files.py
```diff
@@ -12,46 +13,58 @@ def __init__(self, basedir, workdir, datadir, exp_name):
...
+#                 pass
+# #                 print 'Tried to remove some restart files, but number '+str(entry)+' does not exist'
                 
-def keep_only_certain_restart_files_data_dir(exp_object, max_num_files, interval=12):
+def keep_only_certain_restart_files_data_dir(exp_object, max_num_files=None, interval=12):
 
     #        sh.ls(sh.glob(P(self.workdir,'restarts','res_*.cpio'))) #TODO get max_num_files calculated in line, rather than a variable to pass.
 
+        if max_num_files is None:
+            month_list = glob(P(exp_object.datadir,exp_object.expname, 'restarts')+'/res*.tar.gz')
```
penmaher, line 61: On a non-unix server (like if you are running on a local machine for example) would it still create a tar.gz?

src/extra/python/scripts/remove_certain_restart_and_data_files.py
```diff
@@ -62,13 +75,27 @@ def keep_only_certain_restart_files_data_dir(exp_object, max_num_files, interval
         for x in files_to_keep:
                files_to_remove.remove(x) 
 
+        first_to_be_removed = True
+        number_removed = 0
+        number_not_removed = 0
             #Then we remove them.
-        for entry in files_to_remove:
+        for entry in files_to_remove[1:-1]:
```
penmaher, line 82: This would benefit from an assert to make sure there is at least one element to loop over

src/extra/python/scripts/remove_certain_restart_and_data_files.py
```diff
@@ -98,36 +125,18 @@ def keep_only_certain_daily_data_uninterp(exp_object, max_num_files, interval=No
...
-#     exp_name_list = ['annual_mean_ice_princeton_qflux_control_1']#, 'annual_mean_ice_post_princeton_qflux_control_1']
-
-#     exp_name_list = ['annual_mean_ice_post_princeton_fixed_sst_TEST_1', 'annual_mean_ice_princeton_qflux_control_matrix_qflux_1']
-
-#     exp_name_list.extend(['simple_continents_post_princeton_fixed_sst_1'])
-
-#     exp_name_list = ['giant_drag_exp_chai_values_without_dc_bug_latest_1']
-#     exp_name_list = ['aquaplanet_qflux_control_1']
+    # exp_name_list = ['']
+    exp_name_list = glob('/disca/share/sit204/data_from_isca_cpu/*/')
```
penmaher, line 131: could this be a non-hard coded option? If not, could you link at the top of the script so it is easy for a user to open and change.

ntlewis:
Just a quick note on this. In Oxford I'm currently using the -Duse_netCDF3 option when compiling Isca with Intel compilers. This is a temporary snag deriving from the fact that here the netcdf library compatible with Intel 2018 compilers was compiled without netCDF4 enabled. 

I just thought I'd mention that the above won't work with a mkmf template that uses-Duse_netCDF3, as pinterp_utilities.F90 and run_pressure_interp.F90 assume netCDF4 and don't have any #ifdef statements to switch to a 'netCDF3' mode. 

I can get around this by compiling the plevel interpolation stuff with gfortran (for which I can access a compatible netCDF4 library).

PR #112, opened 2020-03-20
Added signature to mosaic_util header file.

aramirezreyes:
Hello:

I am trying to compile Isca to run on the Cori Computer at NERSC (I am a PhD student at UC Davis working with Dr. Da Yang)

When trying to compile using gcc, I get a warning that the function `isHeaderNode()` is not defined when trying to compile mosaic. It is true, and the function is defined `at mosaic_util.c` but it does not appear on the corresponding header file. 

This commits adds the corresponding signature to `mosaic_util.h` and allows to continue the compilation.

Apologies if it is an inappropriate commit.


gkvallis:
Thanks. Certainly not inappropriate.   We’ll take a look.  But it may take a bit longer than normal.

RuthG:
I think this looks good overall, it's nice to have more users and environment options in the code!

As a rule, our main criteria for code changes, aside from that they achieve the desired outcome, is that they don't affect the results of previous work unless there was a bug. I think there might be a couple of files in your pull request that could affect current test cases/model running outside NERSC:
- `held_suarez_test_case.py` : you've commented out the compile step, is this edit meant to be included in this pull request? I think this will cause problems with our trip-tests.
- 'templates/run.sh' : edits here look like they change a current default? 

I'll run trip-tests on this request sometime over the next few days to check the other edits don't have any unexpected side-effects, but at a glance I can't see that they should.

Hope you're having fun with the model, thanks for contributing!

Ruth

RuthG:
Unfortunately these edits failed all of our trip tests, with the model crashing on every one, and so I can't merge this in. 
Thanks for offering this though! If you'd still like to get it merged in please could you look at the minimal changes that would need to be included, and make sure these are included as options that are off as default, so as to avoid issues that could affect set-ups of existing users.
I've added some guidance to the front page README on contributing to Isca, in case that is helpful.
Best wishes,
Ruth

aramirezreyes:
Hi Ruth. Thanks for checking this out. Now I see that the PR is quite dirty, I thought I had included only one minor change. Will make a second attempt soon.

PR #113, opened 2020-04-07
Added ape sst and fixed indenting

penmaher:
Hi team,
This is a straight forward PR. While working on ITCZ-MIP I needed to generate an APE SST profile. I am now looking to merge this into the master.
Penny

### DIFF
--- a/src/atmos_spectral/driver/solo/mixed_layer.F90
+++ b/src/atmos_spectral/driver/solo/mixed_layer.F90
@@ -145,7 +145,7 @@ namelist/mixed_layer_nml/ evaporation, depth, qflux_amp, qflux_width, tconst,&
                               ice_albedo_value, specify_sst_over_ocean_only, &
                               ice_concentration_threshold,                   &
                               add_latent_heat_flux_anom,flux_lhe_anom_file_name,&
-                              flux_lhe_anom_field_name, do_ape_sst
+                              flux_lhe_anom_field_name, do_ape_sst, qflux_field_name
 
 !=================================================================================================================================
 
--- a/src/atmos_spectral/driver/solo/mixed_layer.F90
+++ b/src/atmos_spectral/driver/solo/mixed_layer.F90
@@ -491,7 +491,7 @@ if (do_sc_sst) then
 else
     if (do_ape_sst) then
         ! if using specified sst without land, do not calc the heat capacity
-        do_calc_eff_heat_cap = .false.
+        do_calc_eff_heat_cap = .true. ! BUG testing only - should be .false.
     end if
 endif
 
--- a/src/atmos_spectral/driver/solo/mixed_layer.F90
+++ b/src/atmos_spectral/driver/solo/mixed_layer.F90
@@ -37,7 +37,7 @@ use  field_manager_mod, only: MODEL_ATMOS
 
 use tracer_manager_mod, only: get_tracer_names, get_number_tracers
 
-use      constants_mod, only: HLV, PI, RHO_CP, CP_AIR
+use      constants_mod, only: HLV, PI, RHO_CP, CP_AIR, KELVIN
 
 use   diag_manager_mod, only: register_diag_field, register_static_field, send_data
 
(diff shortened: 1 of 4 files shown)


sit23:
Thanks Penny - would you mind running the trip tests for us to check nothing else has changed? Once you're done, could you post the results here?

src/atmos_spectral/driver/solo/mixed_layer.F90
```diff
@@ -129,10 +130,10 @@ module mixed_layer_mod
                               delta_T, prescribe_initial_dist,albedo_value,  &
                               land_depth,trop_depth,                         &  !mj
                               trop_cap_limit, heat_cap_limit, np_cap_factor, &  !mj
-			                        do_qflux,do_warmpool,        &  !mj
+			                  do_qflux,do_warmpool,              &  !mj
```
dennissergeev, line 133: why is there an indent here?

src/atmos_spectral/driver/solo/mixed_layer.F90
```diff
@@ -671,10 +695,10 @@ subroutine mixed_layer (                                               &
 
     if(do_sc_sst.and.specify_sst_over_ocean_only) then
         where (land_ice_mask) delta_t_surf = - corrected_flux  * dt / eff_heat_capacity
-	where (land_ice_mask) t_surf = t_surf + delta_t_surf			 
+	    where (land_ice_mask) t_surf = t_surf + delta_t_surf
```
dennissergeev, line 698: Here and 3 lines below, why the indent?

src/atmos_spectral/driver/solo/mixed_layer.F90
```diff
@@ -643,21 +646,42 @@ subroutine mixed_layer (                                               &
...
+         where (.not.land_ice_mask) t_surf = t_surf + delta_t_surf			 
 	 else
 	     delta_t_surf = sst_new - t_surf
 	     t_surf = t_surf + delta_t_surf
 	 endif
-	     
 end if
 
-if ((.not.do_sc_sst).or.(do_sc_sst.and.specify_sst_over_ocean_only)) then
+if(do_ape_sst) then
```
dennissergeev, line 663: I think it would be nice to say what "ape" stands for (in my mind it's always available potential energy...)

Also, from a complete novice point of view, I wonder if it'd be good to have some notes about this SST choice somewhere at the level of the user interface / docs...

penmaher:
As a reminder when I forget in the future: syntax is ./filename old_id new_id -r path. The -r path is needed if not using SSH.

./trip_test_command_line b8c07a45835d6e66164588f2a054275533d207a3 cf214c194932ab819c07e7170dd8a4b08d733e60  -r https://github.com/ExeClim/Isca.git

src/atmos_spectral/driver/solo/mixed_layer.F90
penmaher: Thanks Dennis. There were tabs on this line that did not display in my editor. Fixed in the next commit.

src/atmos_spectral/driver/solo/mixed_layer.F90
penmaher: As above

src/atmos_spectral/driver/solo/mixed_layer.F90
penmaher: I have added further description at L663 to say what APE stands for and the reference. Also at the point where the logic flag is declared. In terms of SST choices, @sit23 do you know if we have a standard location for this type of setting info?

penmaher:
I have addressed the comments above and added a test case for the APE set-up. I will add this to the trip tests in the next commit.

penmaher:
The ape_aquaplanet test failed (as expected) as the commit id b8c07a (the current HEAD of the master) does not have output to compare to.  

```Results for all of the test cases ran comparing b8c07a and d2bd15 are as follows...
socrates_aquaplanet : pass
variable_co2_grey : pass
held_suarez : pass
variable_co2_rrtm : pass
bucket_model : pass
top_down_test : pass
frierson : pass
axisymmetric : pass
MiMA : pass
ape_aquaplanet : fail
giant_planet : pass
Nightmare, some tests have failed```

penmaher:
Stephen let me know if you want me to merge your trip test changed into the PR.

sit23:
Also, just to be clear, your commit 078b598 is because `qflux_field_name` was already a variable, but it just wasn't a namelist variable?

sit23:
For this to go through, I think you need to mark your changes as done, @dennissergeev?

penmaher:
Thank you @sit23 . Yes the ```qflux_field_name``` had a default value set but was not an imput.

PR #116, opened 2020-04-27
Modified qflux and mixedlayer files

natgeo-wong:
I have added the following options to the qflux module:
* Additional parameters for gaussian warmpool: `warmpool_lond` and `warmpool_latd` that allow for customization of the size of the gaussian warmpool.
* Because I modified the inputs for the gaussian warmpool somewhat, I also modified the MiMA warmpool submodule a bit for vectorization.

I have not tested the MiMA warmpool modifications, but I have used the Gaussian warmpool module quite often so the Gaussian warmpool should be correct.

### DIFF
--- a/src/atmos_param/qflux/qflux.f90
+++ b/src/atmos_param/qflux/qflux.f90
@@ -10,14 +10,20 @@ implicit none
 real ::    qflux_amp      = 30.,  &
            qflux_width    = 16.,  &
            warmpool_amp   =  5.,  &
-           warmpool_width = 20.  
+           warmpool_width = 20.,  &
+           warmpool_lond  = 20.,  &
+           warmpool_latd  = 20.
+
 integer :: warmpool_k     = 1
 
-logical :: qflux_initialized = .false.
+logical :: qflux_initialized = .false., &
+           warmpool_MiMA     = .false., &
+           warmpool_gaussian = .false.
 
 
-namelist /qflux_nml/ qflux_amp,qflux_width,&
-                     warmpool_amp,warmpool_width,warmpool_k
+namelist /qflux_nml/ qflux_amp,qflux_width,warmpool_amp,&
+                     warmpool_MiMA,warmpool_width,warmpool_k,&
+                     warmpool_gaussian,warmpool_lond,warmpool_latd
 
 private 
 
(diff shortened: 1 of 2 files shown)


RuthG:
Hi Nat,
Going through these now. Looks good, thanks so much for contributing your edits, it's great to have some community input.

First a coding question:  I need to give it some thought as to a) if it matters and b) which I think is 'right', but is there a reason you've changed latb/lonb to (I think) lat/lon here? 

Next a note on procedure: We have some trip tests set up to run for code changes to make sure they don't lead to any surprises. There is documentation on running these here: https://github.com/ExeClim/Isca/blob/77a3d49c5e3131dc6312d32b3698feac2cc8d156/exp/test_cases/trip_test/README.md 
The envs additions can't really change anything, and I'll check these this time round for the other edits, but for future changes if you can run them quickly and confirm the results that'd be great :) 

And last a general style point: We've been having some discussion in the Isca team recently about whether we include names/initials in code and more broadly how we document the code. Going forward, our plan is to clean the code by removing names & initials as we make further changes to modules, as the contributor can be seen via the Github record. So I'm going to scrub your name off these, but it's nothing personal! The commit log will of course show the changes came from you.

natgeo-wong:
Hi Ruth!  Thanks for letting me know about the trip_tests.  I think I might have missed that when I was reading through the README about how to contribute.  I will do that next time!  Thanks for letting me know!

I can't quite remember why I changed to lat/lon, but I remember there was some reason behind it relating to it being a vector or something.  It was also a personal stylistic reason because I think I was trying to merge two codes together from another version of FMS I used to use while working under Tapio.

RuthG:
Ah, good point! I have put in a request to add information on contributing to the front-page README. Most code edits (rather than env additions) have been internal or some via visitors so I don't think we actually had public guidelines on how we test changes previously...

Ok, I'll trip-test and also check if latb-> lat changes the output when this option is set. If it does discuss with others in Exeter. 
Thanks again for putting this forward to merge in!

RuthG:
I've just trip-tested this and it failed on 'MiMA', which is a test case where some of the q-flux code is used. I think this will relate to the latb -> lat change. I'll take a look at the code differences and a think about the best way to resolve this.

natgeo-wong:
I'm guessing the failure is due to the `warmpool` option?  Because I've used the `gaussian` option quite extensively so I don't think there's a problem there.

When is a trip-test considered a fail?  Do you compare the outputs of the two different versions?

RuthG:
The trip-tests compare runs with identical namelists for the model version based on the last commit from the pull request and the last commit from before the pull request (both input by user). If there is any difference in output then that is a fail. Generally making changes to how existing configurations behave is to be avoided, unless it's a bug fix. 
I imagine the changes themselves work and do what they're meant to, but if they cause a change in results compared to previous versions then that'll be a test fail. I'll run these cases out a little longer when I have some time, see what the difference looks like in the long-term, and discuss with others.
Sorry, I don't want to dishearten you from contributing! I know from my own contributions that these tests can be frustrating, but it's important to try to meet that standard so we make sure we keep the codebase performing in a consistent way. I've added some guidance to the front page README for future contributors following our discussions :)

natgeo-wong:
Oh no worries about that, I'll keep those guidelines in mind next time when I'm making modifications to Isca!  I'm just curious as to how the results are being read.

Also, it seems like each of the trip tests spins up for 10 years!  Is that the case?

RuthG:
Ah no, that would take ages! 😁  Each run is just 3 days to quickly show if anything has changed. In fact I think the compilation of RRTM and Socrates took longer than all the runs for me! Runs are done from: https://github.com/ExeClim/Isca/blob/master/exp/test_cases/trip_test/trip_test_functions.py

natgeo-wong:
Alright @RuthG I think I'm going to close this pull-request for now, since I've rebased the Isca code back to the `master` branch of ExeClim for now, and because I've got more familiar with `git` so I want to do this more consistently.  The same for the PRs https://github.com/ExeClim/Isca/pull/119 and https://github.com/ExeClim/Isca/pull/118

PR #119, opened 2020-04-27
Modified the Land-Generator Function and added examples.

natgeo-wong:
Modified Land-Generation function. Examples found in `write_land_example.py` script.

The following features were added:
* I can define a custom path (using `fname`) instead of it being purely dependent on `GFDL_BASE`
* It can write to grids of different resolutions instead of just being purely T42
* I can write multiple custom landmasses instead of just one

### DIFF
--- a/input/land_masks/write_land_example.py
+++ b/input/land_masks/write_land_example.py
@@ -1,0 +1,104 @@
+from isca import write_land
+
+# create an Experiment object to handle the configuration of model parameters
+# and output diagnostics
+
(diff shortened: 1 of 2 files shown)


natgeo-wong:
Thanks for the comments Ruth!

1) Oops I dropped the ball on that one.  Will fix that and repush.  Thanks for catching!
2) This sounds like a better option but I need to go and sit down and think about this some more on how best to implement it.

Will get to this asap.

PR #166, opened 2020-05-15
Check isca python installation is in GFDL_BASE used

RuthG:
Print a warning if the isca python installation used is inconsistent with the GFDL_BASE directory


sit23:
This looks great - thanks Ruth! I wonder if it might be good though to add a little bit more detail to the message as to how this scenario might happen and why it might be a problem? It wouldn’t have to be much, but might help a user sort this problem out.

RuthG:
Good point! I have added a little more detail, but tried to keep it brief.

PR #179, opened 2020-08-04
Fixed line length nml fix

sit23:
The latest version of `f90nml` available on pip, v1.2, enforces fixed-column widths for string variables when writing out nml files. The default line width is 72 characters. The default is the same in previous versions of `f90nml`, but did not seem to be enforced. This has the effect of splitting some variables across multiple lines in `input.nml`, causing the model to fail when these are read. An example case is `socrates_rad_nml` with the `lw_spectral_filename` parameter pointing to a directory that is generally longer than 72 characters. I have modified the python front end to allow for a maximum width of 350. This is longer than the 256 characters this namelist parameter has as its own maximum, therefore leaving room for long variable names too.

### DIFF
--- a/src/extra/python/isca/experiment.py
+++ b/src/extra/python/isca/experiment.py
@@ -144,8 +144,11 @@ class Experiment(Logger, EventEmitter):
     def write_namelist(self, outdir):
         namelist_file = P(outdir, 'input.nml')
         self.log.info('Writing namelist to %r' % namelist_file)
-        #A fixed column width is added to be a fixed number as most string namelist variables are width 256 in Isca, so that width plus some indentation and the namelist parameters own name should not exceed 350 characters. Default f90nml value is 72, which is regularly too short for some namelist variables where directories are pointed to.
-        self.namelist.column_width=350
+        # A fixed column width is added to be a fixed number as most string namelist variables are
+        # width 256 in Isca, so that width plus some indentation and the namelist parameters own 
+        # name should not exceed 350 characters. Default f90nml value is 72, which is regularly 
+        # too short for some namelist variables where directories are pointed to.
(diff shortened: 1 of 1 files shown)


src/extra/python/isca/experiment.py
```diff
@@ -144,6 +144,8 @@ def update_namelist(self, new_vals):
     def write_namelist(self, outdir):
         namelist_file = P(outdir, 'input.nml')
         self.log.info('Writing namelist to %r' % namelist_file)
+        #A fixed column width is added to be a fixed number as most string namelist variables are width 256 in Isca, so that width plus some indentation and the namelist parameters own name should not exceed 350 characters. Default f90nml value is 72, which is regularly too short for some namelist variables where directories are pointed to.
```
dennissergeev, line 147: PEP8: make the comment fit 100 character width (split into several lines) and put a space after `#`

src/extra/python/isca/experiment.py
```diff
@@ -144,6 +144,8 @@ def update_namelist(self, new_vals):
...
+        self.namelist.column_width=350
```
dennissergeev, line 148: PEP8: put spaces around `=`

src/extra/python/isca/experiment.py
sit23: Thanks Denis - I've now done this.

src/extra/python/isca/experiment.py
sit23: Done

sit23:
Thanks for the review @dennissergeev - and you're right that it'd be good to get us conforming to these standards. Hopefully the changes I've just made look alright.

sit23:
Looks like the build has passed OK, so I'll merge this in now.

PR #180, opened 2020-08-12
Fix typo

penmaher:
Fix typo from dt_qg_convection to dt_tg_convection.


penmaher:
@rosscastle fixed typo as discussed.  Sorry for the faff for a typo of this nature.

penmaher:
@dennissergeev  Any idea why the Travis CI failed for this?

dennissergeev:
@penmaher it stalled again. I've restarted it, but I think we need to do something about these builds to prevent them from stalling...

sit23:
@dennissergeev is there perhaps a way to increase the time-out limit? It seems like it often gets stuck compiling the Iscacodebase, which is the one with RRTM in, and that takes much longer to compile than the other codebases. Perhaps we could give it a bit longer to do this (although I don't really know how Travis works!).

dennissergeev:
@sit23 I moved the discussion to #182

PR #186, opened 2020-09-23
Frierson Test Case Fix (Issue 184)

rosscastle:
To fix the frierson test case issue, line lengths in diag_table.f90 a…nd mpp.f90 were changed from 256 to 512 where read_ascii_file is called. Other consequences unknown, this may only be a temporary fix.

### DIFF
--- a/src/shared/diag_manager/diag_table.F90
+++ b/src/shared/diag_manager/diag_table.F90
@@ -357,7 +357,7 @@ CONTAINS
     INTEGER, INTENT(out), OPTIONAL, TARGET :: istat
     CHARACTER(len=*), INTENT(out), OPTIONAL :: err_msg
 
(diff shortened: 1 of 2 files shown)


matthewjhenry (approved):

rosscastle:
Note to stop the current master (dc4fed0) crashing on the trip tests, I had to remove f90nml (v1.1.2) from the Isca environment, and install f90nml==0.23. 

Results for all of the test cases ran comparing dc4fed0 and a87d877 are as follows...
axisymmetric : pass
bucket_model : pass
frierson : pass
giant_planet : pass
held_suarez : pass
MiMA : pass
top_down_test : pass
variable_co2_grey : pass
variable_co2_rrtm : pass

PR #187, opened 2020-09-23
Idealised moist docs

RuthG:
Here are the corrections to my previous pull request. 
Apologies for the new branch, I decided to rename my local branch and this seems to have caused some problems with the remotes.

### DIFF
--- a/docs/source/modules/idealised_moist_phys.rst
+++ b/docs/source/modules/idealised_moist_phys.rst
@@ -28,7 +28,7 @@ The fortran file is found in ``Isca/src/atmos_spectral/driver/solo/idealized_moi
 
 Namelist options
 ----------------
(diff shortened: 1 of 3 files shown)


mp586 (approved):

mp586:
are we supposed to add an Authors section or is that optional?

rosscastle (approved):
Really really nice! 

I think the only issue I can find is that the file itself (and therefore in the index) has idealized spelt with an s (correctly) not a z (as per the model code), but I think we can ignore this!

RuthG:
Thanks for checking it, I've added the authors at the bottom, I forgot about that. Happy to merge?

rosscastle:
> Thanks for checking it, I've added the authors at the bottom, I forgot about that. Happy to merge?

Merge time!

mp586:
yes! :)

PR #190, opened 2020-10-21
Fixed typo in docs idealised_moist_phys.rst

daw538:
I have corrected a small typo relating to the options of the `do_seasonal` parameter in this namelist, switching the true/false statements that were the wrong way around.


rosscastle (approved):
Nice one mate!

PR #193, opened 2020-11-10
Giant Planet Test Case "Bug" fix

rosscastle:
As agreed with ST, after problems spotted by DW, the giant planet tes…t case crashes due to a memory problem. This crash happens on GV5 and ISCA HPC, therefore we assume it happens on most machines. It does not happen on the trip tests because this test case is set to a lower resolution for these. To ensure the test case works out of the box, we increase the stack size (and number of cores, as it was very slow). The original version did say in the fms namelist that the stack size should be increased for high res runs, but there is no error pointing to this.

### DIFF
--- a/exp/test_cases/giant_planet/giant_planet_test_case.py
+++ b/exp/test_cases/giant_planet/giant_planet_test_case.py
@@ -4,7 +4,7 @@ import numpy as np
 
 from isca import IscaCodeBase, DiagTable, Experiment, Namelist, GFDL_BASE
 
-NCORES = 4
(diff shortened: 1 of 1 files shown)


rosscastle:
Test passed for giant_planet. Commit d836edf gives the same answer as commit 6216a21
Results for all of the test cases ran comparing 6216a21 and d836edf are as follows...
giant_planet : pass

sit23 (approved):
Looks good to me - thanks @rosscastle and Dan

PR #198, opened 2020-12-16
Constants Docs (for future constants.f90)

daw538:
This is just a first draft of some documentation for the constants module (to be continued in new year).

Note that this contains a number of new namelist parameters that are not currently present, especially those relating to moist atmospheric parameters. This sets the scene for an updated `constants.f90` (and associated modules) that generalises the atmosphere to more easily allow for a wider range of configurations beyond water-based Earth-like systems, as discussed with @sit23. Modifications to the modules themselves will form part of a future PR, and this should not be accepted until then.

I also suspect that `ptsd` may be surplus to requirements, but I'm not sure if it's still used anywhere.

### DIFF
--- a/docs/source/modules/constants.rst
+++ b/docs/source/modules/constants.rst
@@ -1,0 +1,86 @@
+
+Constants: Atmospheric and Planetary Parameters 
+===============================================
+
+Summary
+-------
+
+``constants.F90`` provides a number of controls that set the fundamental parameters of the planet being studied - it is used by nearly every other module in Isca. By default, Isca models a planet equivalent to earth in terms of physical characteristics such as size, surface gravity and rotational period. In addition, water is set as the default condensate for any moist physics modules that may be used.
+
+Planetary Parameters
+^^^^^^^^^^^^^^^^^^^^
+
+Below are a set of controls that allow the physical size and rotation to be set. This allows us to model a range of planets, including others in the Solar System such as Mars and Jupiter, in addition to exoplanets. Rotation can be set by two different parameters: a rotation rate :math:`\Omega` can be specified; or alternatively a period in seconds can be given that is converted back to a value of :math:`\Omega`.
+
++-------------------+----------------------------------+---------------------------------------+-------------------------------------------------+
+| Name              | Default                          | Units                                 | Description                                     |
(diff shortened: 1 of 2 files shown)


rosscastle:
Nice one chap, happy to review after xmas. You'll need a title line in the index file too for the links to work.

rosscastle (commented):
Really good stuff mate! Great title and summary and useful structure to the document. Very clear and concise. 

A few points:

- In the planetary parameters paragraph I think it would be useful to emphasise once that these are namelist controls which the user can set (I think from then on it's clear in the subsequent paragraphs).
- L13, I would maybe stick with the word ``omega`` for clarity, just because that is what you refer to it as in the table. Out of interest what happens when you specify both omega and the period? Which one wins?
- I know it's true to the code, but why are we limiting ourselves to 1 decimal place with Earths gravity? 9.81 at least :P 
- Generally where you have "Note" I would include the note next to it. E.g. "Note: While..." as it looks a bit like the notes is missing at the moment.
- For the dry atmosphere table cp_air isn't in the constants namelist that I can see, is it in IMP?
- Similarly for cp_vapour. I know you specify that the moist parameters are with the addition of IMP, but maybe make it clearer which it's for, e.g. they don't need to specify the hlv etc in IMP too.
- Can you add in the solar_const, earthday_mutiple and es0 parameters? 
- YY and ZZ to Stephen and me.

Couple of spelling errors:

- Dry air heat capAcity
- e.g. L53 spectral_dynamics_nml namelist, to me this is repeating namelist twice. Not sure if people will know nml on its own? Or just spectral_dynamics namelist.
- L71 no U in vaporisation
- L8 Earth is capital E when using Earth on its own, or lower case when preceded by "the".

daw538:
Given some of the changes made to `constants.f90` seemed to be causing issues when it came to running the trip tests, I have made a shortened version of the docs that omits the moist atmosphere settings, which allows it to be released with the current release of the model.

Stephen and I have discussed a couple of ideas for getting around the issues however it's not high priority currently and I'd rather get some docs out than this getting left behind.

Given the previous comments have been addressed it should be pretty much ready to accept.

rosscastle (approved):
Agreed RE getting the docs out. I'm happy if you are!

PR #199, opened 2020-12-16
Diag_manager_mod documentation PR

mp586:
Documentation for diag_manager_mod focusing mostly on how to save Isca output ( at different frequencies ) and including a list of standard diagnostic variables.

### DIFF
--- a/docs/source/modules/diag_manager_mod.rst
+++ b/docs/source/modules/diag_manager_mod.rst
@@ -1,0 +1,162 @@
+Diagnostics Manager Module & Saving model output 
+==============
+
+Summary
+-------
+
+This module handles the writing of diagnostic output to netCDF files. The user can specify which fields should be output and at which temporal resolution (e.g. monthly means, daily means ... ). The source code is located at ``src/shared/diag_manager/diag_manager.F90``. 
+
+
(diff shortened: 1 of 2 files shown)


rosscastle:
Hi Marianne, sorry this took me so long to get round to! I think this is really great, there's not much I can think to improve really. A few very minor stylistic points:

- I think it could be made a bit clearer how often you can output all the data on multiple timescales (e.g. all data hourly, daily, monthly) for 1 run. I think it would help if you explain what each of the 'module, name, time_avg, files' are - importantly make it clear that 'files' refers to the 'atmos_monthly' etc. Also you could make it clear that you only need 1. In the second sentence of the second paragraph I think it would make more sense if you started with 'If you only wanted specific variables saved on at a particular output frequency(cies), then you can...' 
- I think it would look a little neater if in the text the namelist options (e.g. true/false/none/123) were also formatted. E.g. line 81: ' If ``files`` = None ' could be ' If ``files`` = ``None`` '.
- In the output field table, the unit for relative humidity has a \ (slash) in it.
- Line 57, I think it would be helpful (although hopefully obvious) to point out that examples of a diag script are in those example runscripts near the top.

PR #202, opened 2021-02-04
units correction for ucomp_height

lqxyz:
The units for `ucomp_height` in `spectral_dynamics.F90` should be `m**2/sec` rather than `m**2sec`.


rosscastle (approved):
Looks good to me!

PR #203, opened 2021-02-11
Generalisation of atmospheres for use on non-Earth/exoplanet systems

daw538:
Currently a number of modules within Isca are hard-coded for use on Earth, a throwback to its initial development as the GFDL model. The result is that moist physics processes currently only work for atmospheres with water as the primary condensate, whilst objects that rotate very slowly compared to Earth (e.g. Titan and Venus) are not correctly modelled without modifications to the code.

This PR aims to address these shortcomings by introducing a more generalised form for moist atmospheres, allowing the use of alternative condensates (e.g. methane on Titan), which allows for the modelling of a wider range of planetary systems including exoplanets. This is controlled by new namelist parameters in the `constants.f90` file -- the documentation for which can be found in PR #198 .

Further, a modification is made to allow very slowly rotating objects to be modelled correctly (under the current Isca these systems are falsely modelled as being tidally-locked). This is accompanied by the addition of a new namelist parameter that allows the rotation period to be specified directly (rather than as a rotation rate).


rosscastle (commented):
Really nice fixes, I can't see anything wrong. I have some questions:

1. two_stream_gray_rad.F90 L424: Is it possible (perhaps as a later project) to be able to use the day_in_sec option rather than using an Earth day (86400)?
2. idealised_most_phys.F90 L427-429: Is this the trip test issue you were fixing?

surface_flux is a simple rename, so that's fine. sat_vapour_press fine, took me a second to get round the "double variables" but this does make sense. Same for constants.

Will go back through the docs PR now.

daw538:
Yes, the change to `idealised_moist_phys.F90` was made to fix the issue where the trip tests would fail at the buckets module for any trip test where moist processes were involved, regardless of whether `'bucket':True`had been set. It seemed to be caused by the lack of explicit allocation in the case.

Regarding your comment on `two_stream_gray_rad.F90` L424, I'm not sure if you can but I can investigate further in future. The reason being that I think certain parts of the way time is dealt with in Isca are hard-coded to Earth timings, so switching to `day_in_sec` may actually break things. Not sure if this makes much sense but again perhaps best left for future investigation.

rosscastle:
> Yes, the change to `idealised_moist_phys.F90` was made to fix the issue where the trip tests would fail at the buckets module for any trip test where moist processes were involved, regardless of whether `'bucket':True`had been set. It seemed to be caused by the lack of explicit allocation in the case where water was not being used as the condensate.

Makes sense

> Regarding your comment on `two_stream_gray_rad.F90` L424, I'm not sure if you can but I can investigate further in future. The reason being that I think certain parts of the way time is dealt with in Isca are hard-coded to Earth timings, so switching to `day_in_sec` may actually break things. Not sure if this makes much sense but again perhaps best left for future investigation.

Yeah that makes sense, future project maybe. I feel like it'll be a big one to sort out all the calendars etc.

> One change I'm considering is changing the variable name of `dens_vapor`, which I now feel is a little misleading. It is the density of liquid state (i.e. 1000 for water) and using vapor doesn't seem quite right. When I originally made the variable I hadn't quite considered the need for documentation down the line so I think this would be good to sort to avoid confusion in future! Perhaps `dens_liquid` is a better fit.

Yeah good idea. Or `dens_fluid`, works for gas or liquid.

PR #204, opened 2021-03-02
Updated isca_job.sh script 

rosscastle:
Changed the shell script from moab to slurm as the ISCA HPC has been changed recently and the current is now out of date.

### DIFF
--- a/exp/test_cases/isca_job.sh
+++ b/exp/test_cases/isca_job.sh
@@ -2,7 +2,7 @@
 # in Exeter.  USER and PROJECTNUMBER in the metadata above need to be
 # updated to appropriate values for your experiment.
 # For more information about running on the Exeter cluster, see the wiki
-# here: http://login.isca.ex.ac.uk:8080/wiki/Isca+Guide.pdf
+# here: https://universityofexeteruk.sharepoint.com/sites/ExeterARC
 
 #!/bin/bash
 #SBATCH --export=ALL # export all environment variables to the batch job
--- a/exp/test_cases/isca_job.sh
+++ b/exp/test_cases/isca_job.sh
@@ -7,7 +7,7 @@
 #!/bin/bash
 #SBATCH --export=ALL # export all environment variables to the batch job
 #SBATCH -D . # set working directory to .
-#SBATCH -p ptq # submit to the parallel queue
+#SBATCH -p pq # submit to the parallel queue
 #SBATCH --time=03:00:00 # maximum walltime for the job
 #SBATCH -A Research_Project-PROJECTNUMBER # research project to submit under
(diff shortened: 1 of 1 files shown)


exp/test_cases/isca_job.sh
```diff
@@ -1,17 +1,20 @@
...
-#PBS -l walltime=0:30:00 # Maximum wall time for the job.
-#PBS -A Research_Project-PROJECTNUMBER # research project to submit under.
-#PBS -m e -M USER@exeter.ac.uk # email me at job completion
-
 # This is an example of how to submit an experiment to the queue on the Isca supercomputer
 # in Exeter.  USER and PROJECTNUMBER in the metadata above need to be
 # updated to appropriate values for your experiment.
 # For more information about running on the Exeter cluster, see the wiki
-# here: https://wiki.exeter.ac.uk/display/ISCA/Isca+Wiki
+# here: http://login.isca.ex.ac.uk:8080/wiki/Isca+Guide.pdf
```
matthewjhenry, line 5: This link does not work for me. :/ But the old one does!

exp/test_cases/isca_job.sh
```diff
@@ -1,17 +1,20 @@
...
+
+#!/bin/bash
```
matthewjhenry, line 7: `#!/bin/sh` on mine but that depends on the shell used.

matthewjhenry (changes requested):
Apart from couple of comments, all looks good, thanks for doing this, Ross!:)

exp/test_cases/isca_job.sh
rosscastle: Hmm, I can't get either of the links to work. Safari and chrome block the old one because there is no security certificate? If you got the old one to work shall we revert to that one? I checked the guide they emailed around September and that defo directs you to the new link.

exp/test_cases/isca_job.sh
rosscastle: I can stick a note in to say that /sh is also an option? Or just leave it?

exp/test_cases/isca_job.sh
rosscastle: I did the sensible thing and emailed the isca support team!

matthewjhenry: Try this one: https://universityofexeteruk.sharepoint.com/sites/ExeterARC
It's a series of word documents though.... lol

exp/test_cases/isca_job.sh
matthewjhenry: Ahh, just leave it, people will figure it out. Sorry, I was being knit-picky.

matthewjhenry (approved):

PR #205, opened 2021-03-31
Begginers guide

rosscastle:
First draft of the beginners guide I have been writing. Please suggest any changes.

### DIFF
--- a/docs/source/begginers_guide.rst
+++ b/docs/source/begginers_guide.rst
@@ -1,0 +1,171 @@
+Isca Beginners Guide
+====================
+
+Below is a list of reading and activites that will help you get comfortable using Isca. Be assured that your supervisor/tutor will not expect you to be fluent with this when you start using Isca, but it will help to have an idea of what to expect when you start running the model.
+
+This document is essentially a suggestion of signposts. With this kind of work, self-study and initiative is very important. It is up to you to go and research the topics until you feel comfortable.
+
+Essentials
+-------------
+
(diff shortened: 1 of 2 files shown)


daw538:
Looks pretty good to me, just a couple of suggestions:

Firstly, I think it should be _Beginner's_ guide (not Begginer) :wink:

There are a couple of places in the text where another section is referenced such as (L47)
> In the future it may be useful to have python environments (see 'Conda' in 'Intermediate').

I think it would be good if these were directly linked in the text, which I think you can do with like [this](https://github.com/ExeClim/Isca/blob/dcd2e7aa68af677f5a3570c7573d43787ab08788/docs/source/begginers_guide.rst#conda) (for reference it's done by appending `#<section>` to the page URL).

Would it be worth stating (perhaps in the file transfer section) that it is also possible to view and transfer files using the native file browser if you're using UNIX or OSX, using their built-in functions to connect via SFTP? I'm thinking this might make navigating around and looking at isca's file structure a bit easier for those not familiar with command line (although tbf it's how I normally navigate about anyway).

Also, though I know this will be quite a subjective view, but could we also suggest nano as a possible text editor? It's available on pretty much every unix system and I think offers a useful middle ground between gedit and vim, which I'd say are at opposite ends of the useability/interface scale.

Other than that, nice work!

rosscastle:
Hi @daw538, thanks for your comments, I've sorted everything other than the editor comment as we discussed! I used an internal link on the docs instead of the URL, however this means I had to edit conf.py to allow. @mckimb if you could let me know if you have any comments. Thanks!

PR #208, opened 2021-05-21
Simple clouds

penmaher:
This simple cloud scheme was implemented for using Isca in the second phase of the SPOOKIE model intercomparison. It then went on to form the base for Qun Liu's paper: https://gmd.copernicus.org/articles/14/2801/2021/

@lqxyz: requesting your review for your expertise on this topic and to ensure I don't put anything on the master that will make your merge more difficult.

@RuthG requesting your input (don't need a full review unless you want to) as the test cases for the cloud scheme rely on your new era topography file (which has improved the Gibbs ripples). This file will is not part of this PR.

@sit23 requesting your input (don't need a full review unless you want to) as I change the socrates test cases.


sit23:
Hi @penmaher - thanks for this. I'll take a look at the test case changes, as you requested.

penmaher:
Following Ruth's suggestion i have added the .nc file used in the test cases

src/atmos_param/cloud_simple/cloud_simple.F90
```diff
@@ -0,0 +1,378 @@
...
+    integer, intent(in), dimension(4) :: axes
+
+    integer :: io, stdlog_unit
+
+
+#ifdef INTERNAL_FILE_NML
+       read (input_nml_file, nml=cloud_simple_nml, iostat=io)
+#else
+       if ( file_exist('input.nml') ) then
+          nml_unit = open_namelist_file()
```
lqxyz, line 75: Hi Penny, it looks like you haven't defined `nml_unit` yet, and perhaps define it as an integer variable in line `68`.

src/atmos_param/cloud_simple/cloud_simple.F90
```diff
@@ -0,0 +1,378 @@
...
+          read (nml_unit, cloud_simple_nml, iostat=io)
+          call close_file(nml_unit)
+       endif
+#endif
+    stdlog_unit = stdlog()
+    write(stdlog_unit, cloud_simple_nml)
```
lqxyz, line 81: Hi Penny, could you add some error checks for the namelist so as to find the misspelled namelist? For example:
```
     integer :: io, ierr, nml_unit, stdlog_unit

#ifdef INTERNAL_FILE_NML
    read(input_nml_file, nml=cloud_simple_nml, iostat=io)
    ierr = check_nml_error(io, 'cloud_simple_nml')
#else
    if (file_exist('input.nml')) then
      nml_unit = open_namelist_file()
      ierr = 1
      do while (ierr /= 0)
        read(nml_unit, nml=cloud_simple_nml, iostat=io, end=10)
        ierr = check_nml_error(io, 'cloud_simple_nml')
      enddo
10    call close_file(nml_unit)
    endif
#endif
    stdlog_unit = stdlog()
    write(stdlog_unit, cloud_simple_nml)
```

src/atmos_param/cloud_simple/cloud_simple.F90
```diff
@@ -0,0 +1,378 @@
...
+        frac_liq = 1.0
+    else if (temp < KELVIN-40.0) then
+        ! All ice if temp is below -40C
+        frac_liq = 0.0
+    else           
+        ! linearly interpolate between T=0 and -40C
+        frac_liq = 1.0 - (KELVIN-temp) / 40.0
+    end if 
+
+
```
lqxyz, line 206: Delete one blank line?

lqxyz (commented):
Hi @penmaher, thank you very much for coding the SPOOKIE cloud scheme. I have reviewed the file `cloud_simple.F90` and all the logics look good to me. I just made some comments on the namelist check, some tidy-ups and the possible simplification of the nested loops (this needs more work, so just neglect it if it's not easy). I will review the other files later.

src/atmos_param/socrates/interface/socrates_interface.F90
```diff
@@ -55,10 +55,13 @@ MODULE socrates_interface_mod
...
   INTEGER :: id_soc_surf_flux_lw, id_soc_surf_flux_sw 
+
   INTEGER :: id_soc_surf_flux_lw_down, id_soc_surf_flux_sw_down 
+  INTEGER :: id_soc_surf_flux_lw_clear, id_soc_surf_flux_sw_clear 
+  INTEGER :: id_soc_surf_flux_lw_down_clear, id_soc_surf_flux_sw_down_clear 
   INTEGER :: id_soc_flux_lw, id_soc_flux_sw
-  INTEGER :: id_soc_olr, id_soc_toa_sw
-  INTEGER :: id_soc_toa_sw_down
+  INTEGER :: id_soc_olr, id_soc_toa_sw, id_soc_olr_clear, id_soc_toa_sw_clear
+  INTEGER :: id_soc_toa_sw_down, id_soc_toa_sw_down_clear
```
lqxyz, line 64: Hi @penmaher, thank you very much for adding more outputs about the clear-sky fluxes, which makes the calculation of cloud radiative effect much easier.

lqxyz (commented):
Hi @penmaher, many thanks for adding the clear-sky fluxes in `Socrates` interface, which makes the calculation of cloud radiative effect much easier! Also, the simple cloud and Socrates interfaces in `idealized_moist_phys.F90` look good to me and the file has become cleaner after your clean-up to the old comments.

penmaher (commented):
Hey @Qun. Thanks for the excellent review. I have made these changes but I am now doing local testing to make sure the model still compiles etc. I will push the changes when I am confident I have not broken anything.

penmaher:
@lqxyz the runscripts needed some updating and the code changes now compile and run too. Should be good for you to review my changes.

src/atmos_param/cloud_simple/cloud_simple.F90
```diff
@@ -284,16 +293,15 @@ subroutine calc_cf(q_hum, qsat, cf, cca, rh, simple_rhcrit, rh_min, rh_max)
...
-        cf = (rh - simple_rhcrit ) / ( 1.0 - simple_rhcrit ) 
-
+        cf = (rh - simple_rhcrit ) / (1.0 - simple_rhcrit)
     else
-        cf = (rh - rh_min ) / ( rh_max - rh_min ) 
+        cf = (rh - rh_min) / (rh_max - rh_min) 
     end if 
 
-    cf = MAX( 0.0, MIN( 1.0, cf))
+    cf = MAX(0.0, MIN( 1.0, cf))
```
lqxyz, line 306: Delete the extra space after `MIN(`?

src/atmos_spectral/driver/solo/idealized_moist_phys.F90
```diff
@@ -203,12 +203,12 @@ module idealized_moist_phys_mod
...
-     temp_2m,		       &   ! used for 10m winds and 2m temp
-     u_10m,		           &   ! used for 10m winds and 2m temp
-     v_10m,		           &   ! used for 10m winds and 2m temp
+     ex_del_m,             &   ! used for 10m winds and 2m temp
+     ex_del_h,             &   ! used for 10m winds and 2m temp
+     ex_del_q,             &   ! used for 10m winds and 2m temp
+     temp_2m,              &   ! used for 10m winds and 2m temp
+     u_10m,                &   ! used for 10m winds and 2m temp
+     v_10m,                &   ! used for 10m winds and 2m temp
      q_2m,                 &   ! used for 2m specific humidity
```
lqxyz, line 212: Thank you very much for making the code aligned and much nicer to read.

src/atmos_spectral/driver/solo/idealized_moist_phys.F90
```diff
@@ -586,11 +586,12 @@ subroutine idealized_moist_phys_init(Time, Time_step_in, nhum, rad_lon_2d, rad_l
 !    initialize damping_driver_mod.
       if(do_damping) then
          call pressure_variables(p_half_1d,ln_p_half_1d,pref(1:num_levels),ln_p_full_1d,PSTD_MKS)
-	 pref(num_levels+1) = PSTD_MKS
-         call damping_driver_init (rad_lonb_2d(:,1),rad_latb_2d(1,:), pref(:), get_axis_id(), Time, & 
-!note that in the original this is pref(:,1), which is the full model pressure levels and 
-!the surface pressure at the bottom. There is pref(:2) in this version with 81060 as surface pressure??
-                                sgsmtn)
+     pref(num_levels+1) = PSTD_MKS
```
lqxyz, line 589: More indentation for this line?

src/atmos_spectral/driver/solo/idealized_moist_phys.F90
```diff
@@ -772,7 +773,7 @@ subroutine idealized_moist_phys_init(Time, Time_step_in, nhum, rad_lon_2d, rad_l
 endif
 
    id_rh = register_diag_field ( mod_name, 'rh', &
-	axes(1:3), Time, 'relative humidity', 'percent')
+   axes(1:3), Time, 'relative humidity', 'percent')
```
lqxyz, line 776: Perhaps more indentation for this line?

src/atmos_spectral/driver/solo/idealized_moist_phys.F90
```diff
@@ -1049,15 +1050,15 @@ subroutine idealized_moist_phys(Time, p_half, p_full, z_half, z_full, ug, vg, tg
...
-			                    ex_del_q(:,:),				                &
-			                     temp_2m(:,:),				                &
-			                       u_10m(:,:),				                &
-			                       v_10m(:,:),				                &
+                                ex_del_m(:,:),                              &
+                                ex_del_h(:,:),                              &
+                                ex_del_q(:,:),                              &
+                                 temp_2m(:,:),                              &
+                                   u_10m(:,:),                              &
+                                   v_10m(:,:),                              &
```
lqxyz, line 1058: This looks great to me.

src/atmos_param/cloud_simple/cloud_simple.F90
```diff
@@ -203,7 +214,6 @@ subroutine calc_liq_frac(temp, frac_liq)
         frac_liq = 1.0 - (KELVIN-temp) / 40.0
```
lqxyz, line 216: Perhaps add more spaces around the minus sign? Changing the `(KELVIN-temp)` to `(KELVIN - temp)`?

lqxyz (commented):
Hi @penmaher, thank you very much for the updates to the codes, which makes the code cleaner and easier to read. All the changes look good to me and only several minor changes are still needed.

penmaher:
Thanks for the second review Qun. All suggested changes have been made and pushed

lqxyz:
Hi @penmaher, the changes look good to me, but do you know why the CI test [(job #120.2)](https://travis-ci.com/github/ExeClim/Isca/jobs/517587548#L936) failed? It seems that the `idealized_moist_phys.F90` could not find the `cloud_simple_mod`, but the `cloud_simple.F90` has been specified in [`path_names`](https://github.com/penmaher/Isca/blob/simple_clouds/src/extra/model/socrates/path_names#L14). But it's a little weird as this branch could be compiled smoothly when I test it locally.

dennissergeev:
@lqxyz could you try checking out this branch in a completely new directory and compiling it from scratch locally?

One error I noticed is in the [compile_mppn.sh](https://github.com/penmaher/Isca/blob/simple_clouds/postprocessing/compile_mppn.sh#L13) script. Line 13 should be changed to 
```bash
source $GFDL_ENV
```

penmaher:
Thanks for the suggestion @dennissergeev. I downloaded a new repo of this branch, it compiles but does not run. I am investigating. This is weird as it runs fine in the clone I did the dev on (though I have not cleaned the compile directory).   

In terms of the suggested change to  compile_mppn.sh script. Line 13, I think the full path and filename are needed, not just the filename, or at least it is in the current way I have the environment variables set.

dennissergeev:
> In terms of the suggested change to compile_mppn.sh script. Line 13, I think the full path and filename are needed, not just the filename, or at least it is in the current way I have the environment variables set.

For the automatic tests, `GFDL_ENV` already includes the full path, see https://github.com/penmaher/Isca/blob/simple_clouds/.travis.yml#L12

It could be changed if you think it's not appropriate, but for now just try changing your `GFDL_ENV` to include the path.

sit23:
Hi @penmaher - thanks for this. The socrates test cases look fine to me, and it's nice to have all the options included. 

Did you want to include one of the clouds on cases in the trip tests? It might be a good idea.

I was also wondering about the RRTM modifications you've included here. Did we actually get those to work in the end? And if not, is it worth excluding them from this? 

As a final note, I'm guessing that you'll run the trip tests on this PR once all the changes are in? No point in running them sooner, perhaps.

sit23:
And to respond to the point about the `compile_mppn.sh`, I think the way most of us have the $GFDL_ENV set up means that the full path is required. But you're right @dennissergeev that it might be better to change this. It would be quite a big change though. We could probably compromise where the front-end would check whether the full path or just the GFDL_ENV file exists, and then uses this one?

dennissergeev:
@sit23 OK, I see. To be honest I don't remember why I set it up like that in `.travis.yml`. 
@penmaher or @lqxyz could you change the file accordingly please (lines 12 and 40)?

lqxyz:
Hi @dennissergeev, I have tried to clone this branch to a new directory, and it could compile correctly. And it still works after modifying the `GFDL_ENV` to its full path in `.bashrc`, and changing `source $GFDL_BASE/src/extra/env/$GFDL_ENV` to `source $GFDL_ENV` in `compile_mppn.sh`.

dennissergeev:
@lqxyz thanks for testing this.

Even though this error seems to be unaffected by `$GFDL_ENV`, could you please add a commit with the changes to `.travis.yml` as I described in my previous comment. Then we'll check the travis test log again.

dennissergeev:
So the compilation failed again, but at least the bug with `$GFDL_ENV` is now fixed.
The reason for the compilation error seems to be that `cloud_simple` is not being compiled prior to `idealized_moist_phys`...

dennissergeev:
@lqxyz @penmaher check that you added `cloud_simple` to _all_ path_names:
```bash
# in isca's top-level directory
find . -name path_names -type f -print0 | xargs -0 grep cloud_simple
```

sit23:
I've run the trip tests on the current head of the master (8a715ca) vs the last 'clean' commit on this simple clouds branch (8c39828, i.e. the last commit before Penny started hunting for bugs). The results all pass:
```
Results for all of the test cases ran comparing 8c39828 and 8a715ca are as follows...
axisymmetric : pass
bucket_model : pass
frierson : pass
giant_planet : pass
held_suarez : pass
MiMA : pass
socrates_aquaplanet : pass
top_down_test : pass
variable_co2_grey : pass
variable_co2_rrtm : pass
ape_aquaplanet : pass
```
This is clearly good news, as it means the simple_clouds branch can be merged in cleanly. However, @penmaher did almost exactly the same tests and found failures, so there's still a bit of mystery here. But, on the face of it, the tests are passing. I think if we can revert the commits @penmaher did when looking for bugs then we can happily merge this in.

sit23:
OK - I've reverted all of @penmaher's bug-finding commits, and there are no differences between the current HEAD (e70400c) and her last clean commit 8c39828 (i.e. `git diff e70400c 8c39828` returns nothing). I'll run the trip tests on the current HEAD just to be sure.

sit23:
In re-examining the trip-tests, I think I have identified the reason for @penmaher's trip test failures when I get the trip test as clean. In this p/r the namelist parameters for the `socrates_aquaplanet.py` experiment are changed.  This means that the results from this experiment will change depending on whether it is run with the new or old configuration. Now the trip tests are designed to run each experiment with a set of commits with a consistent namelist. However, the situation can arise where someone runs the trip test with a given commit using a given version of the test cases, but then the test cases are changed, and then the trip tests are run again using one of the same commits as before. In that case, the data from the previous run (using the old namelist) is not overwritten, but the new commit uses the new namelist to run. So the trip tests will appear to fail, but only because they have been run with different namelists. I have therefore added a check to the trip tests to make sure that two experiments that are being compared are run with the same namelist values.

I managed to reproduce one of @penmaher's failed trip tests, and this was the reason for it in my case, so I'm guessing it was the case for @penmaher too. On that basis, I have updated the trip tests in this p/r to make sure this doesn't happen again. I also now believe the trip tests to be clean for this p/r. I'm now running the trip tests for the head of this branch vs the current master. Once those are done and clean, I'm happy to merge this in.

sit23:
OK, here are the results of the trip tests for the current head of the master () and the current head of the simple clouds branch ():
```
Results for all of the test cases ran comparing eaa2dc3 and 8a715ca are as follows...
axisymmetric : pass
bucket_model : pass
frierson : pass
giant_planet : pass
held_suarez : pass
MiMA : pass
socrates_aquaplanet : pass
socrates_aquaplanet_cloud : fail
top_down_test : pass
variable_co2_grey : pass
variable_co2_rrtm : pass
ape_aquaplanet : pass
```
All of the tests pass apart from the `socrates_aquaplanet_cloud` case. But this is because it requires the use of the cloud module, and this is not present in the commit at the current head of the master (), and so that model run fails. I am therefore happy that this code passes the necessary tests, and can be merged in. I will complete my review to allow this to happen.

exp/test_cases/socrates_test/socrates_aquaplanet.py
```diff
@@ -94,7 +91,7 @@
         'chunk_size': 16,
         'use_pressure_interp_for_half_levels':False,
         'tidally_locked':False,
-        #'solday': 90
+        'solday':90
```
sit23, line 94: It is noted that this change will change the result of the test case. But this is fine as it will certainly be an obvious change.

sit23: No action is required on that basis.

sit23 (approved):
All fine as far as I can see. Trip tests are passing. Happy for this to be merged in.

dennissergeev:
Nice! I guess the next step is to add documentation for this scheme (see #147)

sit23:
Agreed @dennissergeev - perhaps this can wait until after @lqxyz has merged his SimCloud code in?

Issue #210, opened 2021-06-07
fail to build on Mac OS (via py.test or otherwise)


gaelforget:
## Description

My attempts to compile Isca on a Mac OS have failed, so for now I give up. Until this issue is fixed (and unless I am just confused) my suggestion would be that you add a warning in the docs & README.md that Isca can only be built on a linux machine (not on Mac).

At first I tried as explained in the README (i.e. via `held_suarez_test_case.py`) and then as explained [in the docs](https://execlim.github.io/Isca/latest/html/install.html) (i.e. via `py.test`). 

I get error messages that have to do with system calls that seem specific for linux:

- `ulimit -s unlimited` in `src/extra/python/isca/templates/compile.sh`
- `syscall(__NR_gettid)` in `src/shared/mpp/affinity.c`
- `cpu_set_t`, `CPU_ZERO`, `coremask`, etc also in `mpp/affinity.c`

The patch below shows what I did to get past the first two errors based on finding [this post](https://github.com/google/glog/issues/185). 

I also found this [other post](https://stackoverflow.com/questions/45227009/alternative-to-shed-getaffinity-cpu-set-t-etc) by @ ntlewis , I think, which links to this [potential solution](http://www.hybridkernel.com/2015/01/18/binding_threads_to_cores_osx.html).

### git diff / patch I tried

```
--- a/src/extra/python/isca/templates/compile.sh
+++ b/src/extra/python/isca/templates/compile.sh
@@ -20,7 +20,7 @@ executable={{ executable_name }}

 netcdf_flags=`nf-config --fflags --flibs`

-ulimit -s unlimited # Set stack size to unlimited
+#ulimit -s unlimited # Set stack size to unlimited
 export MALLOC_CHECK_=0

 # 3. compile the mppncombine tool if it hasn't yet been done.
diff --git a/src/shared/mpp/affinity.c b/src/shared/mpp/affinity.c
index f0eb781..e165365 100644
--- a/src/shared/mpp/affinity.c
+++ b/src/shared/mpp/affinity.c
@@ -30,9 +30,13 @@
 #include <sys/resource.h>
 #include <sys/syscall.h>

+#include <pthread.h>
+
 static pid_t gettid(void)
 {
-  return syscall(__NR_gettid);
+  uint64_t tid64;
+  pthread_threadid_np(NULL, &tid64);
+  return (pid_t)tid64;
 }

 /*
```

gkvallis:
[...]

PR #211, opened 2021-06-11
[WIP] working version on gfortran11 on macOS v11.2

gaelforget:
**I am opening the PR as a way to share and discuss but shouldnt be merged in its current form** 

This follows up on https://github.com/ExeClim/Isca/issues/210 and are modifications which eventually allowed me to compile and run `held_suarez_test_case.py` on a Mac (running OS 11.2). 

There were several types of issues as you'll see, some of which I uncovered after upgrading to the latest gfortran compiler (v11)

- linux system calls which error out on Mac; some I just commented out for now; the affinity part is switched off via a `MACOS` cpp flag; more or less as done by @jamesp  in 2018  (https://github.com/jamesp/Isca/tree/mac)
- the `-fallow-argument-mismatch` and `-fallow-invalid-boz` part is related to warnings that have recently been turned into errors (the options revert to warnings). Would need a switch like in https://github.com/MITgcm/MITgcm/pull/480 probably for gfortran < v10
- in `create_xgrid.c` I needed to add a definition of `isHeadNode_b` (same as in `mosaic_util.c` but for the name)
- added `gfortran_mac` option file

Maybe this could get reviewed and merged after some revision. Please chime in

### DIFF
--- a/src/extra/env/gfortran_mac
+++ b/src/extra/env/gfortran_mac
@@ -1,0 +1,9 @@
+echo Loading basic gfortran environment
+
+# this defaults to ia64, but we will use gfortran, not ifort
+export GFDL_MKMF_TEMPLATE=gfort
+export F90=mpifort
+export CC=mpicc
+export CDEFS=" -DMACOS "
+export NETCDF_LIBS=`nf-config --flibs`
+
--- a/src/extra/python/isca/templates/compile.sh
+++ b/src/extra/python/isca/templates/compile.sh
@@ -18,9 +18,9 @@ template_debug={{ template_dir }}/mkmf.template.debug
 execdir={{ execdir }}        # where code is compiled and executable is created
 executable={{ executable_name }}
 
-netcdf_flags=`nf-config --fflags --flibs`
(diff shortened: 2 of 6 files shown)


ntlewis:
@gaelforget Quick question: do you have an intuition for whether or not the additional changes you've made (i.e., beyond those in Jamie's branch) would affect compilation on Mac setups for which Isca currently does compile (i.e., what Jamie had working)?

gaelforget:
> @gaelforget Quick question: do you have an intuition for whether or not the additional changes you've made (i.e., beyond those in Jamie's branch) would affect compilation on Mac setups for which Isca currently does compile (i.e., what Jamie had working)?

Except for the `-fallow-argument-mismatch` and `-fallow-invalid-boz` part that need to be skipped with gfortran < 10 (see https://github.com/MITgcm/MITgcm/pull/480) I would think so. Would be awesome if you could double check that

jamesp:
Great to see progress being made on this @gaelforget, @ntlewis!

@gaelforget how is the performance when you run on mac hardware?

PR #212, opened 2021-06-16
RAS output problems

sit23:
**The problem**

Recently it was reported that there were issues with the diagnostic outputs from the RAS modules (the scheme itself seems to run fine). The problems were that the output appeared different depending on the number of cores used, and it was clearly non-hemispherically symmetric, even when the setup was symmetric in other senses.

**The fix**
Old 2006 FMS version requires is, js, and sometimes ks to be supplied to send_data. More recent FMS, upon which Isca is based, seems to not want these variables to be passed, and runs better without them. The old syntax was still being used in RAS, so problems were reported with the outputs. This commit corrects this problem, and the outputs are much improved.

### DIFF
--- a/src/atmos_param/ras/ras.f90
+++ b/src/atmos_param/ras/ras.f90
@@ -1159,41 +1159,41 @@ endif
     rain_ev0(:,:)   =  rain_ev0(:,:)   * dtinv
 
      if ( id_tdt_revap > 0 ) then
-        used = send_data ( id_tdt_revap, dtemp_ev0, Time, is, js, 1 )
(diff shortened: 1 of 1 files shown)


sit23:
Trip tests are passing, so I think this is safe to merge in.

```
Results for all of the test cases ran comparing 75880af and 2b603fa are as follows...
axisymmetric : pass
bucket_model : pass
frierson : pass
giant_planet : pass
held_suarez : pass
MiMA : pass
socrates_aquaplanet : pass
top_down_test : pass
variable_co2_grey : pass
variable_co2_rrtm : pass
ape_aquaplanet : pass
```

Issue #48, opened 2018-05-25
Error checking of namelists


sit23:
Not all of the namelists are being error checked when they are read in. This means a nml read-in can fail without alerting the user, meaning the default values are used without the user knowing. 

This is an issue when the compiler option `INTERNAL_FILE_NML` is set, which we are using.

Examples of ones that aren't error checked:
https://github.com/sit23/Isca/blob/master/src/coupler/surface_flux.F90#L816
https://github.com/sit23/Isca/blob/master/src/atmos_spectral/driver/solo/idealized_moist_phys.F90#L306

Examples of ones that are error checked:
https://github.com/sit23/Isca/blob/master/src/atmos_spectral/model/spectral_dynamics.F90#L253

PR #213, opened 2021-06-25
Sorting out check_nml_error calls

sit23:
Many of the fortran modules in Isca rely on reading namelists to set their parameters. The FMS upon which Isca is built has a built-in-method (`check_nml_error`) to make sure that the namelists are read in properly. This is to prevent problems such as when a namelist parameter is put in the namelist for a module that doesn't exist in that module. When this happens and `check_nml_error` is _not_ called, the reading of the namelist for that module fails, but the user is given no warning. The module in question then runs with its default values for all namelist parameters. This is clearly a bad thing, as you want to know when the namelist is not being read. On the other hand, when check_nml_error is run and it finds a problem as just descried, the model throws a fatal error.

As described in issue #48, not all of Isca's modules do this error checking. This P/R aims to add namelist error checking to all remaining fortran files that include namelists. I first wrote a python script to find all of Isca's fortran files, and then find those that do error checking, and those that do not. I then manually went through and added the checks to all the modules it brought up. This should mean that namelist error checking is done my every module that requires it.

The trip tests have been performed for these files, and they pass:
```
Results for all of the test cases ran comparing 2b603fa and 3b170ce are as follows...
axisymmetric : pass
bucket_model : pass
frierson : pass
giant_planet : pass
held_suarez : pass
MiMA : pass
socrates_aquaplanet : pass
top_down_test : pass
variable_co2_grey : pass
variable_co2_rrtm : pass
ape_aquaplanet : pass
```

### DIFF
--- a/src/atmos_spectral/driver/solo/idealized_moist_phys.F90
+++ b/src/atmos_spectral/driver/solo/idealized_moist_phys.F90
@@ -6,7 +6,7 @@ module idealized_moist_phys_mod
   use fms_mod, only: open_namelist_file, close_file
 #endif
 
-use fms_mod, only: write_version_number, file_exist, close_file, stdlog, error_mesg, NOTE, FATAL, read_data, field_size, uppercase, mpp_pe
+use fms_mod, only: write_version_number, file_exist, close_file, stdlog, error_mesg, NOTE, FATAL, read_data, field_size, uppercase, mpp_pe, check_nml_error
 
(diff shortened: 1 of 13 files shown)


penmaher (approved):
Hi Stephen. No changes requested from me. Clear PR purpose, the changes are helpful, changes are straightforward and pass the trip tests. All good to merge in.

sit23:
Thanks @penmaher. Merging now...

PR #214, opened 2021-06-30
Vertical Coordinate Docs

daw538:
First draft of docs page for the `vert_coordinate.f90` module, used to control the model levels used in a run.

Suggestions/improvements welcome, especially if you spot a typo.


docs/source/modules/vert_coordinate.rst
```diff
@@ -0,0 +1,90 @@
...
+        'pk': [0.000000, 0.000000, 0.000000, 0.000000, 0.000000, 0.000000, 0.000000, 0.000000],
+       }
+
+Even Sigma
+^^^^^^^^^^
+This option will create an evenly spaced set of levels from sigma level 0 (top) to 1 (surface), scaled by the surface pressure set in ``spectral_dynamics``. To use this scheme, set ``'vert_coord_option': 'even_sigma'`` and also provide a value for ``num_levels``. Increasing the number of levels will decrease the thickness of each model layer.
+
+Uneven Sigma
+^^^^^^^^^^^^
+The use of evenly-spaced model levels is often not the best option for modelling atmospheres, since it can lead to low resolution near the surface and in the troposphere, where the atmosphere is often the most dynamic). In addition, poor resolution near the model top can in certain configurations be unstable, since breaking gravity waves deposit significant amounts of energy into the upper layers.
```
sit23, line 67: I think there's a point to make in your first sentence here that it's more the other way around. If you split 1000hPa up into 10 levels, you'll have a 200hPa, 100hPa, and 0hPa level, meaning the entire stratosphere disappears into about 2 levels, whereas you'd get the other 8 in the troposphere.

docs/source/modules/vert_coordinate.rst
```diff
@@ -0,0 +1,90 @@
...
+Climate models split the atmosphere into a number of different levels in order to calculate various quantities (e.g. temperature :math:`T`, zonal wind :math:`u`) throughout the depth of the model. There are a number of ways one may wish to divide the atmosphere; this is the purpose of the ``vert_coordinate`` module. Isca provides four different methods for prescribing model levels:
+
+1. Manually prescribed levels 
+2. Even Sigma levels 
+3. Uneven Sigma levels
+4. Hybrid Scheme
+
+which we will elaborate on later, but first a note about model levels in climate models.
+
+Pressure or Sigma Levels??
```
sit23, line 17: I wonder if you'll need to mention here that the vertical coordinate module actually prescribes the values of the _half_ levels and then subsequently calculates the full levels. So normally there's a value with b=1 at the surface, which is a half level. Then there's some way the model calculates the full levels from the half levels. I'm assuming these are the different solid and dashed lines in your figure?

docs/source/modules/vert_coordinate.rst
```diff
@@ -0,0 +1,90 @@
...
+
+Hybrid Scheme
+^^^^^^^^^^^^^
+In models with particularly high top levels (for example models including the stratosphere or of other planets), running with a conventional sigma level scheme can be undesirable since the model can become quite unstable; calculations in the uppermost layers are more stable if the layers are flat. To overcome this we can specify a hybrid model level scheme, which combines :math:`\sigma`-levels in the lower atmosphere with :math:`p`-levels in the upper atmosphere (and a transition region in between).
```
sit23, line 73: Using sigma levels high up 'imprints' the shape of the topography / surface pressure the entire way through the atmosphere, which is undesirable, as you say.

sit23 (changes requested):
This is great Dan, nice work. I've made a few suggestions, but they're relatively minor. One other thing you might say is to note that the output of the model is on the levels you prescribe in this module (i.e. on sigma levels if you ask for sigma), and therefore it needs to be interpolated to standard pressure levels for interpretation. You could then link to any interpolator documentation should this exist already, or point to it being forthcoming.

docs/source/modules/vert_coordinate.rst
daw538: Yes agreed, probably got mixed up with vertical height z

docs/source/modules/vert_coordinate.rst
daw538: Yes the solid and dashed lines represent the half/full levels.

Good point though, probably best to introduce the concept before detailing the different options.

daw538:
Thanks Stephen for the feedback -- think I've manged to address and integrate everything you've suggested into the new version.

Issue #221, opened 2021-08-27
Seg faults and nml errors when model is compiled with gfortran


RuthG:
## Description

Running the test cases when the current master is compiled with gfortran raises a few issues:
axisymmetric crashes with a nml error in both cases
giant_planet produces a different result every time so tests are uninformative
top_down_test crashes with a seg fault
Socrates runs - I've not tested these personally as I've not currently got SOCRATES installed.

I'm raising this issue for discussion of these issues, as users compiling with gfortran may run into trouble. I'll also check the trip tests on ISCA HPC and gv-emps ifort compilers once these are back online, plus check the SOCRATES tests look ok.

-----
**Screenshots**

-----
**Isca version**

Commit ID: 4006f4438df
Working off the master

-----
**System Information:**
To help us track down bugs faster, please provide the details about the machine where you run Isca.

To do this, run the following code in Python:
```py
import platform
print(platform.uname())
```
and paste the output below.
```txt
uname_result(system='Linux', node='baikonur.ex.ac.uk', release='2.6.32-696.23.1.el6.x86_64', version='#1 SMP Tue Mar 13 17:46:31 CDT 2018', machine='x86_64')
```
-----

PR #222, opened 2021-08-27
Small fixes for test cases that fail under gfortran

sit23:
As described in #221, some of the test cases are failing when run under gfortran. I've identified and fixed two of these 3 issues, and am working on the 3rd. The changes are as follows:

* The axisymmetric case was failing because we were supplying dt_rad as a float when it should have been an integer. Intel didn’t mind this, but gfortran objects.
* The Top down test seg-faulted because one of the loops over the y dimension was looping over the length of the x dimension instead
* I’ve had a look at the giant-planet test case to see if I can see anything obvious. I’ve made a few changes that might help, but I’d need to get a working example of what the problem is before I can make too much progress. 

I'm running the trip tests on these now to see what progress I might have made, but as they all failed previously, this is more just to check that they now run.

### DIFF
--- a/exp/test_cases/axisymmetric/axisymmetric_test_case.py
+++ b/exp/test_cases/axisymmetric/axisymmetric_test_case.py
@@ -141,7 +141,7 @@ exp.namelist = namelist = Namelist({
     'rrtm_radiation_nml': {
         'do_read_ozone':True,
         'ozone_file':'ozone_1990',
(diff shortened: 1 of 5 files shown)


RuthG:
Looks good to me!

sit23:
Thanks @RuthG - would you mind doing a review and approving it so that I can merge it in? Thanks!

RuthG (approved):
Changes are minor bug fixes, confirmed not to influence trip tests on intel compilers. I am happy for these to be merged in, thanks for spotting it all Stephen! :)

sit23:
Thanks @RuthG

PR #224, opened 2021-10-07
Bve swe docs

rosscastle:
Documentation page with links to the GDFL pdfs for the Barotropic vorticity equation solver and shallow water equation solver.

Do we need to add anything else?


RuthG (approved):
All looks fine to me

PR #226, opened 2021-10-13
lscale_cond docs

rosscastle:
A very quick overview of lscale_cond. As we discussed in the meeting, I've just pointed to the Frierson 06 paper and mentioned the diagnostics.


RuthG (commented):
I made some small changes but otherwise looks alright to me

RuthG (approved):

PR #227, opened 2021-10-17
Shallow water and Barotropic Vorticity equation code

sit23:
To complement the 3D-dynamics models already available in Isca, I have added the shallow-water and barotropic vorticity equation solvers. The fortran code for these came directly from GFDL, and was downloaded from ftp://ftp.gfdl.noaa.gov/perm/GFDL_pubrelease/Spectral_Idealized_Atmospheric_Models/spectral_idealized_public_release.tar.gz on 14/05/19.

These two new configurations are run in much the same way as a standard Isca configuration, but import the ShallowCodeBase or BarotropicCodeBase rather than the standard IscaCodebase or SocratesCodebase. They come with options to include stirring or not, which in the shallow-water case is an enhancement of the GFDL shallow-water code. In addition, I have added to the shallow-water code the ability to setup the model with 1.5-layers, thus representing some deep flow beneath the active layer. This feature was used in https://rmets.onlinelibrary.wiley.com/doi/full/10.1002/qj.3755 . 

In addition to these two new configurations, I have added a stirred and non-stirred test case for each of the barotropic vorticity and shallow-water models, and have added these to the trip tests to preserve their functionality in future releases. This required some modifications of the trip tests to import the new codebases and to add custom output files for these cases.

I have tested the modifications with the intel fortran compiler and the gfortran compiler. I will post the results of the trip-tests below.


sit23:
I have run the trip-tests separately for the intel and gfortran compilers. Under gfortran, I get the following results:

```
Results for all of the test cases ran comparing 403f230 and a9610f5 are as follows...
axisymmetric : pass
bucket_model : pass
frierson : pass
giant_planet : pass
held_suarez : pass
MiMA : pass
socrates_aquaplanet : fail
socrates_aquaplanet_cloud : fail
top_down_test : fail
variable_co2_grey : pass
variable_co2_rrtm : pass
ape_aquaplanet : fail
barotropic_vort_eq_stirring : fail
shallow_water_stirring : fail
```
There are 5 fails here. 4 of them are because the current state of the master (403f230) has some issues running with gfortran, which are fixed by #222. These are the 3 socrates cases (socrates_aquaplanet, socrates_aquaplanet_cloud and ape_aquaplanet), and the top_down_test. The other 2 fails are for the new test cases added in this p/r (barotropic_vort_eq_stirring and shallow_water_stirring). These fail because the current master doesn't have the code to run them. I am therefore satisfied that this code runs as expected under gfortran.

sit23:
It is also to be noted that I have made some edits to most of the existing test cases in this p/r. These changes are simply to move the `cb.compile()` statement to the main body of each test case. This means that the code is not unnecessarily compiled when the trip tests import the namelists from these scripts, thus making the trip tests much faster.

sit23:
Overlap with #197 is noted and will be checked.

PR #236, opened 2022-05-30
Update mkmf file to fix gfotran compile issue

ntlewis:
Github test does not compile on master due to compiler issue, this should fix

### DIFF
--- a/src/extra/python/isca/templates/mkmf.template.ubuntu_conda
+++ b/src/extra/python/isca/templates/mkmf.template.ubuntu_conda
@@ -18,8 +18,8 @@ NC_LIB=`nc-config --flibs`
 #  -O2: Level 2 speed optimisations
 
 FFLAGS = $(CPPFLAGS) $(NC_LIB) -cpp -fcray-pointer \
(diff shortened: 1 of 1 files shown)


ntlewis:
Ok so it's fixed the first error, but now there's a load of type errors. I'll try to work on this over the next month, although in my opinion it's low priority as I think it's related to the compiler that the tests are using, but not those that we actually use in Exeter to compile the model. Obviously this does need to be fixed at some point.

dennissergeev (approved):
Even though these options are [discouraged](https://gcc.gnu.org/onlinedocs/gfortran/Fortran-Dialect-Options.html) by the GCC devs, I am happy to approve this for now. In the long run, the relevant errors should be fixed (see #237).

sit23 (approved):
Thanks for doing this @ntlewis. As @dennissergeev says, this isn't the ideal fix, but it will do for now. Fixing all those type-conversion errors in particular is going to take a long time. Perhaps something for some RSE time if we ever get it. But for now let's approve this.

Issue #247, opened 2023-03-16
sh 2.0.0 breaks Isca installation


sit23:
## Description
I've just tried to do a fresh install of Isca on maths2 (one of our newer linux servers here in Exeter), and it's installed the `sh` package at version 2.0.0. When I try and run the Held Suarez test case, I get the following error:

```
python held_suarez_test_case.py 
Traceback (most recent call last):
  File "/emmy-noether/home/sit204/Isca/src/extra/python/isca/helpers.py", line 69, in git_run_in_directory
    git_test = codedir_git.log('-1', '--format="%H"').stdout
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'stdout'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/emmy-noether/home/sit204/Isca/exp/test_cases/held_suarez/held_suarez_test_case.py", line 10, in <module>
    cb = DryCodeBase.from_directory(GFDL_BASE)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/emmy-noether/home/sit204/Isca/src/extra/python/isca/codebase.py", line 35, in from_directory
    return cls(directory=directory, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/emmy-noether/home/sit204/Isca/src/extra/python/isca/codebase.py", line 398, in __init__
    super(GreyCodeBase, self).__init__(*args, **kwargs)
  File "/emmy-noether/home/sit204/Isca/src/extra/python/isca/codebase.py", line 86, in __init__
    self.git = git_run_in_directory(GFDL_BASE, self.codedir)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/emmy-noether/home/sit204/Isca/src/extra/python/isca/helpers.py", line 73, in git_run_in_directory
    git_test = codedir_git.log('-1', '--format="%H"').stdout
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'stdout'
```

[...]

PR #255, opened 2023-07-25
Fix for Sh > 2.* issue, and new environment files and template for running on maths2 (Exeter)

sit23:
As described in issue #247, versions of `sh>2.*` have been causing the isca python front-end to fail. This is because `sh>2.*` has changed the default return type of some of its commands, returning `str` types rather than `sh.RunningCommand`. 

This fix adds a new function to check the type of sh outputs and only asks for `.stdout` when a `sh.RunningCommand` type is returned. This should allow the front end to function with versions of `sh=1.*` and `sh=2.*`. 

Also included here are new environment files and template files for running with intel compilers on Exeter's new maths2 machine. They are included here for my own convenience, but should be of use to others.

### DIFF
--- a/src/extra/env/maths2
+++ b/src/extra/env/maths2
@@ -1,0 +1,6 @@
+echo loadmodules for maths2 machines
+
+export F90=mpiifort
+export CC=mpiicc
+
+export GFDL_MKMF_TEMPLATE=maths2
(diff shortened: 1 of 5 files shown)


dawzeki (approved):
Looks sane to me from a read of the code.

Nice to have the env files for the new maths machines!

daw538 (approved):
Changes look good to me from a read of the code.

Nice to have the env files for the new maths machines!

PR #260, opened 2024-01-24
Adding new environment files for stability

sit23:
We have recently run into problems with the latest version of the openmpi package that is installed via conda ('openmpi=5.0.1'), which caused various glibc errors:

`2024-01-24 12:28:12,404 - isca - INFO - /emmy-noether/home/sit204/mambaforge/envs/isca_env_jan_24/bin/../lib/gcc/x86_64-conda-linux-gnu/12.3.0/../../../../x86_64-conda-linux-gnu/bin/ld: /emmy-noether/home/sit204/mambaforge/envs/isca_env_jan_24/lib/libmpi.so: undefined reference to `memcpy@GLIBC_2.14'`

`2024-01-24 12:28:12,406 - isca - INFO - /emmy-noether/home/sit204/mambaforge/envs/isca_env_jan_24/bin/../lib/gcc/x86_64-conda-linux-gnu/12.3.0/../../../../x86_64-conda-linux-gnu/bin/ld: /emmy-noether/home/sit204/mambaforge/envs/isca_env_jan_24/lib/./libpmix.so.2: undefined reference to `clock_gettime@GLIBC_2.17'`

To combat this, I've updated the main environment file to only use openmpi version 4, as this appears not to have this problem. 

As a second item in the p/r, I've added a frozen version of the environment file for python 3.12, as this seems common amoungst other python packages. It is a configuration we have tested, and it seems to work stably. 

The final change here is a move away from python 3.9 to 3.12, and a corresponding change in the test yml file (which is now set to use the frozen version of the environment file).


ntlewis (approved):
Happy for this to go in. With time, it would be better to patch issues introduced by changes in python package versions within the Isca code, but currently this isn't feasible given the time commitments of the Isca developers.

PR #265, opened 2024-04-30
Adding front-end option to compile Isca using single-precision (only working with Intel compilers)

sit23:
We want to be able to run Isca using single-precision arithmetic, rather than the default double precision. This p/r mirrors some code I wrote way back in 2020:
https://github.com/sit23/Isca/commit/efd019566e6c822a93564a3e7371dc2d4c3de404

This change gives a method on the codebase object to allow the compile flags to be set to do single precision arithmetic, rather then double, but leaves double as the default. 

I've added a new held-suarez test case to show how this method is to be used.

N.B. this only works at the moment for Intel compilation.


sit23:
Closed in favour of #309

PR #266, opened 2024-05-31
Add thermodynamic sea ice code 

ntlewis:
Adds thermodynamic sea ice code following [Zhang et al. (2022)](https://agupubs.onlinelibrary.wiley.com/doi/abs/10.1029/2021MS002671). 

Test case can be found here: exp/test_cases/thermodynamic_sea_ice/grey_thermo_ice_test_case.py


ntlewis:
@sit23 @daw538 Would it be possible for one of you guys to take a look at this at some point? Thanks :)

src/atmos_spectral/driver/solo/idealized_moist_phys.F90
```diff
@@ -1336,9 +1347,12 @@ subroutine idealized_moist_phys(Time, p_half, p_full, z_half, z_full, ug, vg, ps
 
 endif ! if(turb) then
 
-! Adding relative humidity calculation so as to allow comparison with Frierson's thesis.
-   call rh_calc (p_full(:,:,:,previous),tg_tmp,qg_tmp,RH)
-   if(id_rh >0) used = send_data(id_rh, RH*100., Time)
+!s Adding relative humidity calculation so as to allow comparison with Frierson's thesis.
+   
+   if(id_rh >0) then 
+    call rh_calc (p_full(:,:,:,previous),tg_tmp,qg_tmp,RH)
```
daw538, line 1353: Was the function call meant to be inside the if statement all along then??

daw538:
Broadly looks fine to me. Most of the code is adding functionality rather than editing the existing model so happy to okay if it works as intended.

PR #277, opened 2025-03-18
Modifications to get Isca to run on Derecho at NCAR

roryck:
These are settings and modifications to build Isca on Derecho at NCAR, using cray-mpich and the Intel compiler suite. There are templates, python environment, and Readme specific to Derecho. Also 2 minor changes in the code
1. Adding a declaration for `isHeadNode(...)` to `src/shared/mosaic/mosaic_util.h`
2. Adding an `#ifdef HAVE_GETTID` block in `src/shared/mpp/affinity.c` as the `gettid(...)` function is supplied in `unistd.h`


roryck:
Glad to hear that helped!

PR #284, opened 2025-08-14
Simple Spectral Longwave Radiation Model

AndrewILWilliams:
Implements the simple spectral longwave radiation model (SSM) into Isca. Paper documenting the scheme is here: https://arxiv.org/abs/2508.09353

Haven't added a trip test yet, and not sure how this would interact with the thermodynamic sea-ice scheme? I might need to edit it a bit if the sea-ice code goes in first.

It would be nice to be able to call the SSM every N timesteps, like RRTMG, but I couldn't work out how to do that because the RRTMG code was confusing to me. If anyone understands that portion of the code and can help out, that would be great!

Thanks
Andrew

### DIFF
--- a/exp/test_cases/frierson/frierson_test_case_ssm.py
+++ b/exp/test_cases/frierson/frierson_test_case_ssm.py
@@ -1,0 +1,194 @@
+import os
+
+import numpy as np
+
+from isca import IscaCodeBase, DiagTable, Experiment, Namelist, GFDL_BASE
+
+NCORES = 8
+base_dir = os.path.dirname(os.path.realpath(__file__))
+# a CodeBase can be a directory on the computer,
+# useful for iterative development
+cb = IscaCodeBase.from_directory(GFDL_BASE)
+
+# or it can point to a specific git repo and commit id.
+# This method should ensure future, independent, reproducibility of results.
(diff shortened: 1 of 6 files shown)


ntlewis:
Thanks Andrew. I will try to take a look at this within the next couple of weeks. Great work!

AndrewILWilliams:
@gkvallis , thanks! And yes, all of the above. Definitely more understandable and easy to modify. 

Re: accuracy, an underappreciated facet of correlated-k schemes is that they involve a fair amount of hand tuning in order to achieve high levels of accuracy (e.g., Section 3.1.2 of Pincus et al., 2019). One could certainly design a four-band, correlated-k scheme which is reasonably accurate, but it would suffer the same "clarity" inadequacies as other correlated-k schemes.

AndrewILWilliams:
Thanks Neil!

ntlewis:
For what it's worth, I think the 'correct' way to resolve the issue above would be to compute some sort of definite integral of the plank function in each bin (and then for the two edge bins, extend the integral to nu=0 and nu=infinity). I will think on how this could be implemented.

AndrewILWilliams:
Hey neil, I've made a copy of my code for reproducibility. Let me know if I should do anything else

AndrewILWilliams:
@ntlewis any chance we could merge this soon? Or you could push your changes?

PR #292, opened 2026-01-30
fix : cg and mg drag mistakenly swapped in a couple places (in the docs)

malcolmmaas:


ntlewis (commented):
Looks good. Thanks for catching this.

PR #301, opened 2026-08-11
Automatically downloads and creates path_names lists for different Socrates versions using new Socrates github

sit23:
One of the significant problems with Isca's use of Socrates is that it has required the user to manually download Socrates themselves, and also to create a list of path_names. This P/R changes both - it uses the Met Office's new github repo for Socrates to automatically checkout Socrates if it is not already present, and then works out which path_names should be used for that version. This was built off the old socrates_versions branch of my fork, with new functionality built and tested by Claude. Claude has done a significant amount of testing of this new functionalityy, but this P/R will remain a draft until I have manually tested the functionality.


sit23:
CI is now passing, and this now includes (for the first time) Socrates-specific compile tests, with the Socrates code downloaded from Github as part of the CI. Going to do some manual testing and then will convert to non-draft P/R.

PR #302, opened 2026-08-13
Merging in Emily Ball's Isca-Mars developments

sit23:
As discussed in #299, it has been my intention for a while to merge Emily Ball's Isca-Mars version into the main Isca. This version was forked from Isca in 2019, and built off my earlier Mars configurations, but now including the option to have latent heating from CO2 condensation, and also a representation of Martian dust using Socrates and prescribed dust fields. I have used Claude to cherry-pick commits from Emily's version to implement this into the main version of Isca, and I have begun testing this version to make sure it doesn't break existing functionality and so that the Mars dust test case works as expected.


sit23:
Updating on the trip test results (Claude's summary).

## Trip test results so far

Comparing this branch (`327dd8c1`) against `master` (`d1321ac3`).

| Test case | Result |
|---|---|
| `socrates_aquaplanet` | ✅ bit-identical |
| `socrates_aquaplanet_cloud` | ✅ bit-identical |
| `ape_aquaplanet` | ✅ bit-identical |
| `grey_mars` | ✅ bit-identical |
| `socrates_mars` | ✅ bit-identical |
| `socrates_mars_dust` | ✅ runs cleanly (new capability, no master equivalent to diff against) |
| `radiative_eq_mars` | ❌ fails — known issue, see below, not caused by this branch |

## Two real fixes found along the way

**Socrates aerosol effects were on for every run, not just Mars dust.** The
dust-forcing port had flipped `control%l_aerosol` unconditionally to `.TRUE.`
in `read_control.F90` (master hardcodes it `.FALSE.`), which broke every
plain Socrates run — crashing with a saturation-vapor-pressure overflow, a
downstream symptom of Socrates' aerosol code choking on species the standard
`ga7` spectral file defines but never configures a parametrization for.
Fixed by gating on `do_dust_forcing` (the namelist flag for the actual
feature) instead of switching it on for every run.

**Mars trip tests crashed on run length.** FMS's `main_nml` days/seconds
units always assume Earth's fixed 86400s/day. The Mars test cases use a
calendar day of 88440s (`dt_atmos=110s`) so an integer number of days fits a
Mars year — 86400 isn't a multiple of 110, so the trip test framework's
generic short-run override crashed on every Mars case. Added a Mars-specific
override (3×88440s) to the trip test driver only; no changes to master.

## Known failing case: `radiative_eq_mars` (not a regression from this branch)

Fails identically on **both** master and this branch with an "unknown
namelist variable" error — five keys in
`radiative_eq_mars_test_case.py` (`do_newtonian_cooling_as_rad`,
`newt_relax_surface`, `update_sst_from_fluxes`, `rh_target`, `delta_t_relax`)
aren't declared anywhere in current master's Fortran.

These almost certainly belong to `top_down_with_moisture` (last touched
2020-04-24, ~979 commits behind master) — the branch this test case's physics
originates from — and were lost when this test case was merged without that
branch ever being reconciled. Deleting the keys locally would make the test
pass but paper over the real gap, so that's deliberately **not** done here.

**Plan:** bring a modernized `top_down_with_moisture` up to date against
master on its own branch/PR, then use that to properly restore
`radiative_eq_mars`. Until then this failure is expected and acknowledged,
not something this PR is trying to fix.

sit23:
Have now run a full trip test suite with these changes. Results (summarised by Claude) are below:

## Full trip test suite results

Comparing `mars_dust_lh` (`63e2f3b2`) against `master` (`d1321ac3`) — all 19
test cases in `trip_test_command_line`'s default suite, plus `socrates_mars`
and `socrates_mars_dust` (excluded from the default suite only because they
need external data files, per their `input/README.md`).

### Summary

| Result | Count | Test cases |
|---|---|---|
| ✅ Bit-identical to master | 18 | `bucket_model`, `frierson`, `held_suarez`, `MiMA`, `realistic_continents_fixed_sst`, `realistic_continents_variable_qflux`, `socrates_aquaplanet`, `socrates_aquaplanet_cloud`, `top_down_test`, `variable_co2_grey`, `variable_co2_rrtm`, `ape_aquaplanet`, `barotropic_vort_eq_stirring`, `shallow_water_stirring`, `column_test`, `grey_mars`, `socrates_mars` |
| ✅ Runs cleanly (new capability, no master equivalent) | 1 | `socrates_mars_dust` |
| ❌ Pre-existing failure, unrelated to this branch | 2 | `axisymmetric`, `giant_planet` |
| ❌ Known, acknowledged failure | 1 | `radiative_eq_mars` |

**Every test case this branch could plausibly affect is bit-identical to
master.** The only failures are either pre-existing on master itself, or an
already-acknowledged gap being tracked separately.

### Pre-existing failures (not caused by this branch)

Both of these crash identically on a plain `master` build, before the
feature-branch side of the comparison is even run:

- **`axisymmetric`**: `FATAL from PE 14/15: interpolator_2D : You must call interpolator_init before calling interpolator`
- **`giant_planet`**: `FATAL from PE 0: spectral_dynamics: temperatures out of valid range`

Neither is touched by anything in this branch's diff. Flagging for
visibility rather than fixing here, since they're out of scope for this PR.

### Known, acknowledged failure: `radiative_eq_mars`

Fails identically on both master and this branch with an "unknown namelist
variable" error. `radiative_eq_mars_test_case.py` sets five namelist keys
(`do_newtonian_cooling_as_rad`, `newt_relax_surface`, `update_sst_from_fluxes`,
`rh_target`, `delta_t_relax`) that don't exist as declared Fortran namelist
variables anywhere in current master. These likely belong to
`top_down_with_moisture` (last touched 2020-04-24, ~979 commits behind
master) — the branch this test case's physics originates from — and were
lost when this test case was merged without that branch ever being
reconciled.

**Plan:** bring a modernized `top_down_with_moisture` up to date against
master on its own branch/PR, then use that to properly restore
`radiative_eq_mars`, rather than papering over the gap with local namelist
deletions. Until then this failure is expected.

sit23:
Given that all the trip tests are passing, bar the known fails, this is now ready to be merged. This means that all the Isca-Mars funcionality is now available within the main version of Isca.

